PR #1, opened 2023-07-18
Adding "Math Addons"

LarsIsHere:
Ok hear me out, I know theres already an extension for operators and for more math blocks, but these blocks literally dont exist like this or at all.  I'll be adding more and more functionality for a variety of blocks. It looks like little amount of blocks but theres a lot of dropdowns
also excuse my bad coding, and 
i have to think about the thumbnail cause this likely might not get added due to being blocks most people wont need.. but u never know
image

Description would be: "Additional math blocks and booleans with constants, trigonometric ratios and other things
Made by LarsIsHere (@larsishere)"

### DIFF
--- a/static/extensions/LarsIsHere/mathaddons.js
+++ b/static/extensions/LarsIsHere/mathaddons.js
@@ -1,0 +1,405 @@
+(function(Scratch) {
+    'use strict';
+
+    class mathaddons {
+        getInfo() {
+            return {
+              id: 'mathaddons',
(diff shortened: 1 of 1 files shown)


Ianyourgod:
I think () root of () is better then your current thing. Also most of these do exist.

LarsIsHere:
aw okay

LarsIsHere:
too many similar blocks

PR #3, opened 2023-07-26
Create ListsPlus.js

pooiod:
Lists+ takes your Scratch projects to a whole new level by introducing nested lists, allowing you to organize information in a more complex and dynamic way. Similar to how you might have folders within folders on your computer, nested lists act like boxes inside boxes, providing a powerful way to manage data efficiently.

image

### DIFF
--- a/static/extensions/pooiod7/ListsPlus.js
+++ b/static/extensions/pooiod7/ListsPlus.js
@@ -1,0 +1,285 @@
+//ListsPlus by scratch.mit.edu/users/pooiod7
+
+class ListPlus {
+  getInfo() {
+    return {
+      id: 'listsPlus',
+      name: 'Lists+',
+      color1: '#ff661a',
+      color2: '#cc5215',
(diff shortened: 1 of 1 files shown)


JeremyGamer13:
putting an invalid list seems to just make the block run forever, maybe you could make it just pretend like an empty list was used if the list is invalid?

also ideally the extension ID should have your name somewhere in it

SharkPool-SP:
array blocks
IMG_6214

RedMan13:
@pooiod :trollface:

JeremyGamer13 (changes requested):
this PR still hasnt been updated, so i'd like to leave it with a change request to mark that there is a reason this isnt merged yet

-putting an invalid list seems to just make the blocks break, maybe you could make it just pretend like an empty list was used if the list is invalid?
- ideally the extension ID should have your name somewhere in it
- the thumbnail may be misleading for some people

SharkPool-SP:
this has gone stale

PR #9, opened 2023-08-22
Create better_storage.js

G1nX-01:
works like the pm storage extension, but with better servers and some more features
for now, it only supports adding/modifying keys to the global servers, but this may change in the future.
servers are much faster than pm's storage, also there's no resetting

everything is encrypted btw lol

there's a feature to lock certain keys. users can only do that by submitting a form
im going to add the ability for the server to not make changes if the key or value is a swear word/slur, probably in the next few days

anyway that's basically it

### DIFF
--- a/static/extensions/Gen1x/better_storage.js
+++ b/static/extensions/Gen1x/better_storage.js
@@ -57,6 +57,11 @@ class Storage {
           text: 'log in first',
         },
         {
+          opcode: 'loggedbool',
+          blockType: Scratch.BlockType.BOOLEAN,
+          text: 'logged in?',
+        },
+        {
           opcode: 'errorlog',
           blockType: Scratch.BlockType.REPORTER,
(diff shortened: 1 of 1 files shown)


JeremyGamer13:
- if the key `scratch` is locked, then why is it the default value for the save key block? ideally you should pick a key that cant be reserved for that
- getting a key that doesnt exist still returns an error, but doesnt set the <server error> or error text blocks to say that there was an error

G1nX-01:
> * if the key `scratch` is locked, then why is it the default value for the save key block? ideally you should pick a key that cant be reserved for that
> * getting a key that doesnt exist still returns an error, but doesnt set the  or error text blocks to say that there was an error

working on that today, will notify here or on discord when i fix those issues
also i will add my name to the id

G1nX-01:
everything's done now

PR #13, opened 2023-09-17
weather ext

G1nX-01:
this extension has 1 simple objective: return weather data.
please note this is a beta and more weather data will be added soon
you cant return the coordinates from the "grab coordinates" block in any possible way cuz security

### DIFF
--- a/src/lib/extensions.js
+++ b/src/lib/extensions.js
@@ -21,6 +21,14 @@ export default [
         isGitHub: true,
     },
     {
+        name: "Sprite Parenting",
+        description: "Link sprites together and make them follow the parent.",
+        code: "SharkPool/Sprite-Linking.js",
+        banner: "SharkPool/Sprite-Linking.svg",
+        creator: "SharkPool-SP",
+        isGitHub: true,
+    },
+    {
         name: "Particle Tools",
         description: "Tools for making particle engines easier.",
         code: "SharkPool/Particle-Tools.js",
@@ -34,6 +42,22 @@ export default [
         code: "Gen1x/better_storage.js",
         banner: "Gen1x/placeholder-betterstorage.png",
         creator: "G1nX",
+    },
+    {
+        name: "Sound Waves",
+        description: "Make sounds with oscillators!",
+        code: "SharkPool/Sound-Waves.js",
+        banner: "SharkPool/Sound-Waves.svg",
+        creator: "SharkPool-SP",
+        isGitHub: true,
+    },
+    {
+        name: "Recording",
+        description: "Record your voice while you run your projects!",
+        code: "SharkPool/Recording.js",
+        banner: "SharkPool/Recording.svg",
+        creator: "SharkPool-SP",
+        isGitHub: true,
     },
     {
         name: "CORS Proxy",
--- a/static/extensions/SharkPool/Recording.js
+++ b/static/extensions/SharkPool/Recording.js
@@ -153,6 +153,7 @@
     recordingSet(args) {
       if (args.MODE === "enabled") {
         if (!warningSent) {
+          // jg: TODO: this should be a Scratch permission.
           const confirmed = window.confirm(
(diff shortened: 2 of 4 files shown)


JeremyGamer13:
- blocks throw an error if location is not set, would probably be better to default the location at latitude and longitude 0
- extension does not use `canGeolocate` on the `grab coordinates` block to make sure the user is ok with their location being used
- most blocks use `console.log` after they are run which could potentially be an annoyance
- other extensions can access the coordinates retrieved from `grab coordinates` via `vm.runtime.ext_weatherext`

G1nX-01:
i have NO idea how to fix the last issue, could i please get some help here?

RedMan13:
> > * blocks throw an error if location is not set, would probably be better to default the location at latitude and longitude 0
> > * extension does not use `canGeolocate` on the `grab coordinates` block to make sure the user is ok with their location being used
> > * most blocks use `console.log` after they are run which could potentially be an annoyance
> > * other extensions can access the coordinates retrieved from `grab coordinates` via `vm.runtime.ext_weatherext`
> 
> i have NO idea how to fix the last issue, could i please get some help here?

me when 
```js
if (!this.longitude) return
if (!this.latitude) return
```
or even
```js
  constructor() {
    this.latitude = 0;
    this.longitude = 0;
  }
```

G1nX-01:
> > > * blocks throw an error if location is not set, would probably be better to default the location at latitude and longitude 0
> > > * extension does not use `canGeolocate` on the `grab coordinates` block to make sure the user is ok with their location being used
> > > * most blocks use `console.log` after they are run which could potentially be an annoyance
> > > * other extensions can access the coordinates retrieved from `grab coordinates` via `vm.runtime.ext_weatherext`
> > 
> > 
> > i have NO idea how to fix the last issue, could i please get some help here?
> 
> me when
> 
> ```js
> if (!this.longitude) return
> if (!this.latitude) return
> ```
> 
> or even
> 
> ```js
>   constructor() {
>     this.latitude = 0;
>     this.longitude = 0;
>   }
> ```

i meant the issue that made the extensions able to get the coordinates

RedMan13:
> > > > * blocks throw an error if location is not set, would probably be better to default the location at latitude and longitude 0
> > > > * extension does not use `canGeolocate` on the `grab coordinates` block to make sure the user is ok with their location being used
> > > > * most blocks use `console.log` after they are run which could potentially be an annoyance
> > > > * other extensions can access the coordinates retrieved from `grab coordinates` via `vm.runtime.ext_weatherext`
> > > 
> > > 
> > > i have NO idea how to fix the last issue, could i please get some help here?
> > 
> > 
> > me when
> > ```js
> > if (!this.longitude) return
> > if (!this.latitude) return
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or even
> > ```js
> >   constructor() {
> >     this.latitude = 0;
> >     this.longitude = 0;
> >   }
> > ```
> 
> i meant the issue that made the extensions able to get the coordinates

i see

RedMan13:
> > * blocks throw an error if location is not set, would probably be better to default the location at latitude and longitude 0
> > * extension does not use `canGeolocate` on the `grab coordinates` block to make sure the user is ok with their location being used
> > * most blocks use `console.log` after they are run which could potentially be an annoyance
> > * other extensions can access the coordinates retrieved from `grab coordinates` via `vm.runtime.ext_weatherext`
> 
> i have NO idea how to fix the last issue, could i please get some help here?

you could make a object inside the extension's env instead of inside the class and store the long and lat there

JeremyGamer13:
do you plan on still working on this? im still open to adding this if the changes do get made eventually

Ianyourgod:
the only time that java has something good: private variables

RedMan13:
> the only time that java has something good: private variables

me when your mom:

JeremyGamer13:
what? i didnt merge that

derpygamer2142:
haxed?

PR #14, opened 2023-09-21
Create storage.js

kokofixcomputers:


JeremyGamer13:
this is an exact duplicate of better storage except with a different url

derpygamer2142:
100/4.2 the extension of all time, should be merged immediately imo

kokofixcomputers:
> this is an exact duplicate of better storage except with a different url

kind of... but with a complete diffrent server

PR #28, opened 2023-10-13
Add CloudLink v0.1.0

MikeDev101:
This PR adds CloudLink to PenguinMod extensions. As far as I'm aware, PenguinMod extensions should be synchronized with upstream TurboWarp, so there is only one minor modification to the upstream CloudLink code: version metadata.


MikeDev101:
Converted back to draft, why do extensions load sandboxed?

MikeDev101:
I have been informed that it is only sandboxed in previews. Marking as ready for review.

PR #35, opened 2023-10-15
Do something with Pause Utilities... (Read Me)

SharkPool-SP:
Posting the file here as theres not really a point in having this as a separate extension as it makes more sense to put it in the Runtime Extension.

Lmk what you think devs!


JeremyGamer13:
because of the nature of this pull request, i'll close it & then add it into runtime extension (runtime is in another repo entirely)
ill only add pause script blocks once i think about a nice way to make pause & unpause project useful

SharkPool-SP:
> because of the nature of this pull request, i'll close it & then add it into runtime extension (runtime is in another repo entirely)
> ill only add pause script blocks once i think about a nice way to make pause & unpause project useful

ok :)

PR #47, opened 2023-10-21
Create howmanylines.js

Monochromasity:
Adds a "How Many Lines in ()" block to the editor. Useful for finding the amount of lines in a user-edited variable.

### DIFF
--- a/static/extensions/Monochromasity/howmanylines.js
+++ b/static/extensions/Monochromasity/howmanylines.js
@@ -1,0 +1,38 @@
+// Name: How Many Lines?
+// ID: monochromasityhml
+// Description: Adds a "How Many Lines in ()" block to the editor. Useful for finding the amount of lines in a user-edited variable.
+// By: Monochromasity
+
+(function (Scratch) {
+  "use strict";
+
+  class HML {
+    getInfo() {
+      return {
+        id: 'monochromasityhml',
+        name: 'How Many Lines?',
+        blocks: [
+          {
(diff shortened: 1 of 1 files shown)


JeremyGamer13:
the only thing i can tell is that you should surround `args.INP` in `Scratch.Cast.toString(args.INP)` to make sure it is always a string

Monochromasity:
> the only thing i can tell is that you should surround `args.INP` in `Scratch.Cast.toString(args.INP)` to make sure it is always a string

okay, i'll add that in. after i do that will it get added very soon or will it take a while?

JeremyGamer13:
sorry for the late reply (we've been busy in the discord & we havent gotten more people who can merge stuff yet)
i think it might be important for some users to mention that the block only counts non-empty lines, maybe renaming the block to 
"how many non-empty lines in ()" could work

Monochromasity:
> sorry for the late reply (we've been busy in the discord & we havent gotten more people who can merge stuff yet)
> i think it might be important for some users to mention that the block only counts non-empty lines, maybe renaming the block to
> "how many non-empty lines in ()" could work

Well an empty line isn't really a line is it? anyway i'll do that later

Monochromasity:
Updated

JeremyGamer13:
this is a good solution, ill add this now

JeremyGamer13:
i only mentioned the empty lines thing since some people might use this combined with other blocks to try and read each line (which would break if the empty lines werent counted)

PR #49, opened 2023-10-23
Pen+ Compat Update

David-Orangemoon:


David-Orangemoon:
The update done. I'm going to start working on collisions and extension based sorting next.

David-Orangemoon:
Fixed some bugs.

JeremyGamer13:
- i cant tell if this is work in progress or not since it keeps changing
- a block's opcode is changed which could corrupt projects
- im not sure how i feel about the addons being used here, might be better if i just made a proper way to do this

David-Orangemoon:
> * i cant tell if this is work in progress or not since it keeps changing
> 
>     * a block's opcode is changed which could corrupt projects
> 
>     * im not sure how i feel about the addons being used here, might be better if i just made a proper way to do this

what block. also I'll remove the addons thing.
I was also just making slight adjustments. But other than that its done!

David-Orangemoon:
Shoot I did something

David-Orangemoon:
@JeremyGamer13 It's finished now.

David-Orangemoon:
Also the op code changes where due to two blocks having overlapping op codes.

PR #54, opened 2023-11-03
Object extension

Skyhigh173:
add object extension, which can process large json files at a fastest rate, from 15x (write) up to 30000x (read) compared to json ext *lol*


Skyhigh173:
There's bug in PM while this is normal in Turbowarp

JeremyGamer13:
> There's bug in PM while this is normal in Turbowarp

what is the bug

Skyhigh173:
> > There's bug in PM while this is normal in Turbowarp
> 
> 
> 
> what is the bug

Try to load this extension into PM

Skyhigh173:
Perhaps the "bug" is the sandboxed issue

JeremyGamer13:
nope, PM automatically casts inputs to their ArgumentType which seems to break this extension
image

Skyhigh173:
lol

Skyhigh173:
@JeremyGamer13 my suggestion is to add `Scratch.ArgumentType.Any` so it won't cast value

JeremyGamer13:
add that to the argument objects, the ones with `type` or `defaultValue`

PR #68, opened 2023-11-23
Add Socket.IO extension.

TechGigant:
image


RedMan13:
you cant import external modules inside the extension
if an external module is necessary it must be built into the extension file and can not ever be imported from an external library

TechGigant:
I have tried that before but exactly that didn't work.

RedMan13:
well you still can not import it externally so

TechGigant:
Why?

RedMan13:
it is also a much better practice as it ensures that nomatter what happens to the service you will always have the code for the extension to work

TechGigant:
Ok well this extension isn't possible then.

PR #85, opened 2023-12-16
Create OpenLink for the pinguinmod extensions

SergeyScratch:
there are two blocks in the extension, the first block opens the link in a new tab, the second block opens the link in a separate window

### DIFF
--- a/static/extensions/OpenLink for the pinguinmod extensions
+++ b/static/extensions/OpenLink for the pinguinmod extensions
@@ -1,0 +1,45 @@
+class OpenLinkExtension {
+    getInfo() {
+        return {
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
So.... network extension. That already exists

TheShovel:
This could be a security risk. You need to use the security manager before opening any page.

RedMan13:
so like, extensions are meant to be .js files

JeremyGamer13:
- file should end with .js
- use the Promise result of `Scratch.canOpenWindow()` to make sure you can actually open the new tab

SharkPool-SP:
this has gone stale

PR #86, opened 2023-12-16
Grayscale Extension

nmsderp:
Small extension that creates a grayscale effect on the entire stage, could be used in like a dramatic point in a game lol. It is just a simple toggle block for the effect.


SharkPool-SP:
This already exists

SharkPool-SP:
Through canvas effects or sprite effects

TheShovel:
This already exists in multiple extensions...

PR #92, opened 2023-12-26
Create images.js

AlphaTechNinja:
this is my image manipulating library 

images

## description ##
this library is used to manipulate different types of images including pngs, jpegs, and svg

 ## features ##
allows writing and reading data from an image
allows creating and importing images as well as exporting (svg is not allowed as it would not make sense mostly because of writing pixels makes rectangles so the data would be really long)

## notes ##
bitmap and svg are currently not supported (even though I added bitmap it doesn't work)

## updates ##
might add common svg functions (aka drawing strokes and shapes)

Next thing to add is cropping and resizing (possibly rotating)

### DIFF
--- a/src/lib/extensions.js
+++ b/src/lib/extensions.js
@@ -240,4 +240,12 @@ export default [
         creator: "MrRedstonia",
         isGitHub: true,
     },
+    {
+        name: "Images",
+        description: "Manipulate images on the spot!",
+        code: "AlphaTechNinja/images.js",
+        banner: "AlphaTechNinja/images.svg",
+        creator: "AlphaTechNinja",
+        isGitHub: true,
+    },
 ];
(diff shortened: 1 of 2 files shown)


AlphaTechNinja:
forgot to add this extension doesn't need to be unsandboxed

AlphaTechNinja:
@SharkPool-SP , @TheShovel , @JeremyGamer13  .Can someone review my extension?, it has been about a week.

RedMan13:
boowomp

SharkPool-SP:
> @SharkPool-SP , @TheShovel , @JeremyGamer13  .Can someone review my extension?, it has been about a week.

Don't rush stuff. Give it time

SharkPool-SP:
You need to add an extension thumbnail and edit the extension.js file to include ur extension. Please do it in this pr and not separate ones.

RedMan13:
you infact do not need to do that and instead just have us do it for you although it wouldnt be bad for you to add it on your own :bleh

AlphaTechNinja:
> You need to add an extension thumbnail and edit the extension.js file to include ur extension. Please do it in this pr and not separate ones.

I thought the thumbnail / banner was optional... anyways I can create a thumbnail if needed

RedMan13:
...it is optional

AlphaTechNinja:
Also the reason I asked is because most of the extensions got reviewed in less than a week

RedMan13:
idk

RedMan13:
seems like its more usefull as an image converter then just general image work in its current state

AlphaTechNinja:
Ok I’m probably just gonna close this and add all the features like svg.  it’s name is images because it isn’t specifically for modifying images but it is one of the key features

AlphaTechNinja:
@SharkPool-SP , @RedMan13 . I am gonna add it to the website so I can see the preview (:

SharkPool-SP:
> ...it is optional

Yeah but it will look like shit without it 🤷‍♂️

RedMan13:
> > ...it is optional
> 
> Yeah but it will look like shit without it 🤷‍♂️

still optional

Baykedy:
what is that thing

AlphaTechNinja:
> what is that thing

The extension?

Baykedy:
> > what is that thing
> 
> The extension?

ok

JeremyGamer13:
sorry that this was abandoned for a while
- the defaultValue for `import image` is way too long, just make it `data:` to entail a data url should be there
- certain blocks tell you that an image isnt selected as an error, but some like `export image` just throw a canvas error instead
- blocks like `import image` dont wait for the importing to finish, which could cause timing problems

RedMan13:
i would think detect the error then either throw a new error to the user that actually states whats wrong or eat the error and act as if nothing is wrong

AlphaTechNinja:
another option I was considering was adding new blocks called `hasErrored` and `errorMessage` but for now it spits out -1 on any block that returns a value. I'm currently working on making import run async so it halts the current script

RedMan13:
add `async` to it???

RedMan13:
or, since this is an event we are waiting for, use the js promise api

AlphaTechNinja:
I was looking and images do in fact have a way to bind a function to the loading of an image and it is possible to use a promise. In this case, I could make the function create a promise that imports the image and resolves on the `image.loaded` method

RedMan13:
```js
new Promise((resolve, reject) => {
    const image = new Image()
    image.onload = () => resolve(image)
    image.onerror = reject
    image.src = 'src url'
})
```

AlphaTechNinja:
ok so I uploaded a new version that waits for the image to import but I still have to test it

AlphaTechNinja:
hold on I got to fix it the promise isn't working right

AlphaTechNinja:
> ```js
> new Promise((resolve, reject) => {
>     const image = new Image()
>     image.onload = () => resolve(image)
>     image.onerror = reject
>     image.src = 'src url'
> })
> ```

penguinmod seems to not add the image to the list so I will try something else (btw I'm still trying this snippet first)

AlphaTechNinja:
might have to scrap the promise idea as it seems to not want to save the image to the `saved` JSON table in PenguinMod

RedMan13:
> > ```js
> > new Promise((resolve, reject) => {
> >     const image = new Image()
> >     image.onload = () => resolve(image)
> >     image.onerror = reject
> >     image.src = 'src url'
> > })
> > ```
> 
> penguinmod seems to not add the image to the list so I will try something else (btw I'm still trying this snippet first)

wadya mean

AlphaTechNinja:
> > > ```js
> > > new Promise((resolve, reject) => {
> > >     const image = new Image()
> > >     image.onload = () => resolve(image)
> > >     image.onerror = reject
> > >     image.src = 'src url'
> > > })
> > > ```
> > 
> > 
> > penguinmod seems to not add the image to the list so I will try something else (btw I'm still trying this snippet first)
> 
> wadya mean

I mean it doesn't register it into the saved list of images

RedMan13:
the code you quoted was never meant to?

AlphaTechNinja:
> the code you quoted was never meant to?

what's happening is I have it to where it adds the image to the `saved` list but for some reason, it doesn't add it.
this is the code chunk of what I wrote (using your promise idea)
```JavaScript
let image = new Image();
saved[name] = image;
return new Promise((resolve, reject) => {
      image.onload = resolve;
      image.onerror = reject;
      image.src = uri;
});
```

AlphaTechNinja:
I might try making an image using the document but I don't think that will make a difference.

RedMan13:
what do you mean "doesnt add it", because thats just not true taken literally, the item gets added to the saved object in the code you provided

RedMan13:
> uri and name exists in the arguments passed to the block. saved is global in the extension so you can never test it without running the extension with the code if you want to try it maybe download it, modify it, and try it on PenguinMod/TurboWarp

what you said makes no sense. all i did was fabricate an instance for the code that was inside the block that works as the block inteads it to work

SharkPool-SP:
Marking as a draft due:
-to Conflicts with other files
-lack of updates/commits from creator

RedMan13:
i mean if its finnished you wouldnt update it right :trollface:

SharkPool-SP:
> i mean if its finnished you wouldnt update it right :trollface:

its not though. There are still unsolved issues

SharkPool-SP:
this has gone stale

PR #109, opened 2024-01-07
Create pangapi.js

oc9x97:


SharkPool-SP:
Little tip, for fetching you should store the last fetched value. If the block is ran with the same arguments, it won't have to fetch anything.

Causes less stress on the api and is faster.

You shouldn't do this for every block

NexusKitten (approved):
There's definitely a lot that can be approved/expanded with this extension, but since it's all functional I don't see a reason to leave it in the pull requests.

PR #112, opened 2024-01-10
Update Sprite Effects & Fix Speech Bubbles & Recording

SharkPool-SP:
### Sprite Effects:
-Code Simplification
-Minor Bug Patches

- New Blend types
- New Advanced Color Map Block

### Speech Bubbles
-Bug Fix


NexusKitten:
Seems legit

PR #120, opened 2024-01-18
Add Scope Variables

Ashimee:
uh it only works in the compiler 😅 .
also I was to lazy to compress and upload so: https://youtu.be/ahXraBAQ2mE

(5 years later I got the thumbnail to work)


NexusKitten (approved):
Not really sure how/if this works but seems legit

PR #126, opened 2024-02-09
Randomly Blocks

Mariocraft987:
No goof blocks and added .md


SharkPool-SP:
We can't merge this. 

You're not allowed to delete blocks. Just hide them

Mariocraft987:
oh

SharkPool-SP:
You're still removing the blocks. Won't merge

PR #135, opened 2024-02-17
Add unicodeConverisons Extension

bruh-9000:
Convert a `String` to `Unicode`, and `Unicode` to a `String`


SharkPool-SP:
This already exists in Operators Expansion. No point in this.

bruh-9000:
Oh damn didn't know that, rip

RedMan13:
yeah so like, pretty sure js strings are already Unicode, being that they are utf-8 encoded iirc

PR #143, opened 2024-02-24
more types extension

lego7set:
it adds 7 new types: object, array, set, map, symbol, nothing, and function (ignore the regexp type i was going to make).

before you say that json and object extensions exists, i have to say that these values are compiled, and are passed by reference instead of value.

please merge.

also tell me if there are any errors or anything i should add.


mariocraft987:
Surely get added 👍

mariocraft987:
https://community-extensions.vercel.app/

lego7set:
nice

RedMan13:
I don't see how passing an id to the object is faster then the objects extension which literally passes the whole objects instead of reparsing everything like the json extension does

RedMan13:
also should be noted: as this isnt meant to be used in turbowarp there is an entire set of APIs built into pm specifically for custom values that can be used here to implement custom renders in monitors snd reporters and also ensure that when parsing/serializing the values get put into existence 
and to be clear I do not mean this needs to changed, just that this can be done instead if desired as the systems do exist

lego7set:
wait there's an entire set of API's for that? i thought my objects were just getting stringified? also they arent ids thrown around, more like literal javascript objects and generator functions cuz i used the compiler.

RedMan13:
no, the objects extension literally just extends the map class to setup some helper functions and a custom stringify function then using that custom map class directly without regeneration

RedMan13:
sorry I missread

RedMan13:
but, on a completely rewriten note that is literally what the object extension does too, as return values aren't casted in any way

also I'ma like actually fully review this in a bit to merge cause :trollface:

RedMan13:
yeah this looks good to me

lego7set:
nice.

mariocraft987:
It does not work!!: https://studio.penguinmod.com/editor.html?extension=https://extensions.penguinmod.com/extensions/VeryGoodScratch42/More-Types.js

PR #162, opened 2024-03-15
Update Boxed Physics image

pooiod:
Moved from png to svg because it's higher quality.


RedMan13:
bruh why cant i delete the image from the web view

RedMan13:
> no longer valid???

you said you where going to do it

PR #163, opened 2024-03-17
add OOP to More Types extension

lego7set:
uhhh i saw the OOP extension from live tests and decided to reimplement it in more types.

this update adds classes, methods, this, instanceof block and a version of the for key value block for tempVars (cuz scratch variables boring)

### DIFF
--- a/src/lib/Documentation/More-Types.md
+++ b/src/lib/Documentation/More-Types.md
@@ -189,9 +189,14 @@ Like functions, methods can execute code and return a value.
 
 ### Construct
 ```Scratch
-Construct an instance of [Insert Class Here]
+Construct an instance of [Insert Class Here]::#B300FF
 ```
 This block creates an instance of the class, and executes the class on the instance.
+### instanceof
+```Scratch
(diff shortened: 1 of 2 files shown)


JeremyGamer13:
does this also fix the bug where the project doesnt save with the blocks

lego7set:
what do you mean? do the blocks disappear when the project is saved?

lego7set:
or are you talking about the fact that the new values from this extension aren't saved?

JeremyGamer13:
i shouldve worded that better
basically the project just cant save if a variable is set to a map for example
image

lego7set:
huh? i dont see this happening?

lego7set:
is this an uploaded project? cuz i havent tested that

JeremyGamer13:
oh this seems to be fixed in the changes you made, sorry if i didnt make it clear

lego7set:
:ok:

JeremyGamer13:
this seems good to merge although we are getting to the point where its difficult for me to tell how to even test this extension :skull:

PR #188, opened 2024-04-28
Change models

cicerorph:


JeremyGamer13:
@TheShovel fyi this broke compat but dont undo this time because its very minor 🤷

PR #195, opened 2024-05-15
V1.9 | PenguinGPT Update

cicerorph:
New features:

> Vision (Make AI see images, only works from gpt-4 and up)
> Claude 3 and LLama support and more models

New blocks:

> Get current model
> Attach image () to next message

### DIFF
--- a/static/extensions/MubiLop/penguingpt.js
+++ b/static/extensions/MubiLop/penguingpt.js
@@ -12,17 +12,21 @@
 
 (function(Scratch) {
     'use strict';
-    console.log("Loaded PenguinGPT v1.7 by MubiLop (https://mubi.tech/)")
+    console.log("Loaded PenguinGPT v1.9 by MubiLop (https://mubi.tech/)")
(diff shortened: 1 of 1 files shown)


RedMan13:
and mis was tested and works?

cicerorph:
yes


> and this was tested and works?

RedMan13:
small stylistic issue: labels do not need (and should not have) opcodes on them

RedMan13:
opps forgot to mergeg :trollface:

cicerorph:
Please, before saying it's not working, verify why its not on the Discord Server of PenguinMod.
The reason why it's not working, its because im rewritting the entirety of the API of PenguinGPT, so please wait.

PuzzlingGGG:
maybe make the blocks temporarily say that it's down for maintenance

RedMan13:
make a stand in server that actually serves an error message to the requestomg client instead of using error server does not exist

cicerorph:
> make a stand in server that actually serves an error message to the requesting client instead of using error server does not exist

Yea i was planning to do that long ago but im so busy i cant do that rn

PR #198, opened 2024-05-17
Add Dictation and WindowHasher extensions

pooiod:
An extension to allow you to use your speech as text
It mostly only works with Chrome though, and also Safari somehow

it comes with 4 blocks:
image
Dictation

Also Window Hasher


RedMan13:
mmmmaby, @JeremyGamer13 whab tdo yo think

JeremyGamer13:
- speech is spelled as "speach" in some parts
- this api is [not widely supported,](https://caniuse.com/speech-recognition) you might want to add a message somewhere about that

pooiod:
Does the "can use" block count?

static/extensions/pooiod/Dictation.js
```diff
@@ -0,0 +1,88 @@
...
+  'use strict';
+
+  if (!Scratch.extensions.unsandboxed) {
+    throw new Error('Your microphone can\'t be accessed in the sandbox');
+  }
+
+  class dictation {
+    constructor() {
+      this.recognizedSpeech;
+      this.cando = Scratch.vm.runtime.isPackaged;
```
SharkPool-SP, line 13: Dont use `runtime.isPackaged` in this use.

Use `typeof scaffolding === "undefined"`. It returns true if its not packaged

JeremyGamer13:
> Does the "can use" block count? Also I fixed the spelling

some browsers have the class but dont fire the events apparently, but i think that'll work fine enough if you dont want to figure out which browsers support it completely

static/extensions/pooiod/Dictation.js
RedMan13: why do that? isPackaged should just be yes if packaged runtime no if unpackaged?

JeremyGamer13: packaged runtime can be disabled in the packager (its very confusing)

SharkPool-SP: > why do that? isPackaged should just be yes if packaged runtime no if unpackaged?

Long story short, isPackaged is kind of misinformation. It doesn't report true if it's in a packaged environment. It returns true depending on certain packager information. 

For example, if you disable the "Remove Assets for RAM" option, isPackaged will return false

pooiod:
I accidentally added WindowHasher to this pull request, but I'm committing to it
WindowHasher
image

RedMan13:
2harts window hasher

SharkPool-SP:
what is window hashing supposed to do?

dumorando:
> 2harts window hasher

the thing after # in a url i thinnk

SharkPool-SP:
fix the conflicts and ill fix up the rest and merge it

CodeFoxy-Github:
i tryed the window hasher it might can't be exploit so lgtm

static/extensions/pooiod/Dictation.js
```diff
@@ -0,0 +1,88 @@
...
+          },
+        ]
+      };
+    }
+
+    WaitAndrecognizeSpeech() {
+      return this.recognizeSpeech(false);
+    }
+
+    canuse() {
```
CodeFoxy-Github, line 52: **How about can use are checked in imported**Edit: bruh no

PR #201, opened 2024-05-25
Fixed the search bar being cut off on most phones

minidogg:


minidogg:
> doesnt happen on my phone? iphone se

on my s22 when i go to the current extra extensions gallery the search bar is cut off unless i flip to portrait.

PR #203, opened 2024-06-07
fix ttsr finally

PuzzlingGGG:
also sorry for the shit indents i literally just through it in a js beutifier lmao cuz it looked like shit


SharkPool-SP:
You shouldn't delete pre existing blocks as it will break projects using them. It's best to just hide them using "hideFromPalette:true"

SharkPool-SP:
Oh wait nvm, GitHub is lying to me with changes made

PuzzlingGGG:
:troll_hands: but uh this pr makes the extension actually functional 💀💀

PR #209, opened 2024-07-15
Change PenguinGPT to PenguinAI

cicerorph:
(Don't pull yet, also i am doing an pull request on the editor aswell so the ext doesnt break there)


cicerorph:
Ok can pull now

JeremyGamer13:
do not rename extension files, it will break compat

JeremyGamer13 (changes requested):
files are renamed & the extension id is renamed, neither of those should be changed

cicerorph:
Fixed

JeremyGamer13 (approved):
menu value change is apparently because old one didnt work before so this is good

PR #212, opened 2024-07-21
Add DogeisCut/Vector2math Extension

DogeisCut:
I've created an extension that adds a completely new vector2 type to Scratch, as well as their associated math functions.

Here are the blocks:
image

Square blocks return vector2s, and empty slots take in vector2s.
Vector2s can be stored in variables and modified remotely via the two command blocks.

Here is the icon:
Vector2Math

Some of these blocks do kind of exist within the 3D Math extension, however, if you are making a 2d project, the 3D Math extension would not be a good choice as you'd only be using a fourth of the blocks, plus this extension doesn't rely on strings to function and uses and actual Vector2 type. Not only that, but this contains more 2D specific math functions involving vectors.


DogeisCut:
> why does it need to be unsandboxed

PenguinMod gives me a strange error every time I try to make it sandboxed, I've even checked with TurboWarp and it worked there, so I'm not sure why it needs to be unsandboxed, but it sure does...

image

RedMan13:
that is weird
well unsandboxed is massively more efficient and gallery extensions are always unsandboxed so it's no big deal

DogeisCut:
Actually, this is still in dev. Dont merge this.

PR #214, opened 2024-07-24
Create updateFile.js

Anonymous-cat1:
Submission
(this is the second time I'm doing this because I had to delete my old ass fork)

**Note: requires to be unsandboxed to function.**

Description
A simple extension that uses the File System Access API to update files dynamically.

Thumbnail


SharkPool-SP:
It would be easier if some of these blocks were simplified into a menu. Also, I don't need 200 comments describing each line

Anonymous-cat1:
> It would be easier if some of these blocks were simplified into a menu. Also, I don't need 200 comments describing each line

OK, those should be addressed now.

SharkPool-SP:
> > It would be easier if some of these blocks were simplified into a menu. Also, I don't need 200 comments describing each line
> 
> 
> 
> (yes you d actually need all those comments)

No. You don't need a comment on every function like telling me what it does

RedMan13:
> No. You don't need a comment on every function like telling me what it does

false
you need comments on all functions so you can just look at it and see "ah yes that's for this" without having to lookup the function yourself and parse it's code into a concept of function

Anonymous-cat1:
y'all what do you want me to do lmao 😭

RedMan13:
> isnt that the point of a function name....

an overtly verbose function name yes, like "drawSquareAtXYWithRotationAndWidthHeight"

but function names like that are *horrible* so you instead write "drawSquare" and append a comment explaining all of its function

this is also why and what jsdoc is and why ts even works at debugging issues, since rather then needing to guess, or even look it up if you have a good editor, you can just go look at the function in its usage and see what its function is, and ts will also *enforce* those descriptions, ensuring you have not forgotten anything before allowing it to actually go and get tested

RedMan13:
> y'all what do you want me to do for it to get added lmao 😭

do he changes that make it not utterly broken :trollface:

Anonymous-cat1:
> > y'all what do you want me to do for it to get added lmao 😭
> 
> do he changes that make it not utterly broken :trollface:

Could you elaborate? It should be working fine on supported browsers (e.g Chrome).

RedMan13:
idk I haven't tested it, only sharkpool has

Anonymous-cat1:
I feel like for this extension specifically, comments aren't really needed because the functions are small and self-explanatory like "getFileContent" or "getFileMetadata". I'm sure they would be needed on larger extensions though

SharkPool-SP:
exactly

SharkPool-SP:
anyways, ill fix this up tonight and merge

PR #218, opened 2024-08-02
Background Remover

DumzDev:


Gen1x-ALT:
I ASKED + BASED + COOL + YOU ARE AWESOME + NOT SKILL ISSUE + W + GOOD EXTENSION

RedMan13:
yeah it should be using something else that will be a bit more solid.........

DumzDev:
> yeah it should be using something else that will be a bit more solid.........

theres a block to set the api key to whatever you want

CodeFoxy-Github:
lgtm

static/extensions/dumzdev/removebg.js
```diff
@@ -0,0 +1,95 @@
...
+                }
+            ],
+            menus: { },
+            color1: '#FFC536', // Main theme color
+            color2: '#FFD06B', // Lighter shade of #FFC536
+            color3: '#FFC536',
+            menuIconURI: 'https://files.catbox.moe/ezlx70.svg', // Menu icon
+            blockIconURI: 'https://files.catbox.moe/ezlx70.svg', // Block icon
+            textColor: '#000000', // Black text on the blocks
+            docsURI: 'https://www.remove.bg/api' // Button linking to Remove.bg API
```
SharkPool-SP, line 47: These should be before the blocks, not after.

Also the block and menu icons should be a data.uri not a direct url.

SharkPool-SP:
You should also remove those comments from the colors, that's useless.

Also also maybe add a label that just warns that it won't work if the api key expires

static/extensions/dumzdev/removebg.js
RedMan13: now the data uri thing is a hill I will die on over this cause needing to paste the raw image data into a JSON argument is terrible

DumzDev: done :)

PR #222, opened 2024-08-11
New extension: Upload 

CodeFoxy-Github:
# Easily Upload file/data to the network 
# features 
image

### upload data (plain text) (catbox.moe)
image
### upload data (plain text) (file.io) (no resolve)
image
### upload data (plain text) (file.io) (with resolve)
image
## open it 
image
image
image
image


SharkPool-SP:
blhe

PR #223, opened 2024-08-12
Cloud Position

SkyBuilder1717:
Extension, that will gets advanced location info about users through Cloud Servers.

**Extension must to be unsandboxed to run.**

### Extension adds:
- _get ip address_ block
- _get location_ block
- _get country info_ block
and other..


SharkPool-SP:
Closing this for 2 reasons:
-there's countless IP grabber extensions, +1 already on the extra gallery
-you gave us no functional code

PR #226, opened 2024-08-14
Update Box2D.js bugfix and math fuctions

pooiod:
I fixed a bug where the [Define polygon from costume] block errors when the sprite is hidden, and added some useful math functions

Screenshot_20240814_180121

Also added dropdown for [GetBodyAt] block to select dynamic or static bodies


CodeFoxy-Github (approved):
Looks Good!

PR #249, opened 2024-11-07
Update penguingpt.js

MiloDev123:
Auto updates Api URL
New "Is Model [] working?" block


cicerorph:
I allow Abby (Milo) to change the contents of my extension.

PR #253, opened 2024-11-19
Vectors extension

CoolDevXD:
This extension add a lot of block for 3D and 2D graphics with vectors. it includes block for cull 2d and 3d vector math camera and sprite movement with vectors and more!


Monochromasity:
gonna check it out

CoolDevXD:
yeah no shit

Monochromasity:
imo cool concept. blocks arent very buggy, seems like a new idea not done before. just wait for someone to try it and then you have a high chance of this getting added i think

> how long does it typically take for an extension to get onto the extension list?

when i tried to get my first extension (and only extension as of now) added, it took about 3 days for someone to tell me about a bug that i then fixed, then about 1-2 more days for it to be added (as far as i can remember) so i would recommend trying to iron out any small bugs/make some small improvements while you wait

PuzzlingGGG:
This seems pretty good but I would recommend acing the id be something like “(your name)vectors” just in case. So something like “cooldev_vectors”

CoolDevXD:
i did

RedMan13:
also no you didn't, only one of the four is technically resolved

CoolDevXD:
> also no you didn't, only one of the four is technically resolved

this is like my first or second time, i have no idea what to do or how this works.

CoolDevXD (commented):
Reviewed changes

RedMan13:
eoic fail, i missed a prior review and assumed none-exisfent

RedMan13:
> Reviewed changes

none of the raised issues have been addressed 🗣️

PR #256, opened 2024-12-04
Integers Extension

xFN10x:
this extension is a simple one, that you can use to work with really small, or large integers.


SharkPool-SP:
Alot of these blocks already exist in other extensions, and some of these blocks are plain useless.

PR #260, opened 2024-12-07
Music to numbers

orengdog:
Hello, I made an extension that turns written numbers into music, could you add it?


orengdog:
Music to numbers
Снимок экрана 2024-12-06 193219

static/extensions/orengdog/Music to numbers.js
```diff
@@ -0,0 +1,323 @@
...
+                id: "musicDigitGenerator",
```
JeremyGamer13, line 93: can you add your name to the ID

static/extensions/orengdog/Music to numbers.js
```diff
@@ -0,0 +1,323 @@
...
+        isPlaying: false,
+        instrument: "sine",
+        instrumentID: 0,
+        instrumentName: "without instrument",
+        pauseDuration: 0.5
+    };
+
+    class MusicGenerator {
+        constructor() {
+            this.audioContext = new (window.AudioContext || window.webkitAudioContext)();
```
JeremyGamer13, line 16: make sure to register this audioContext to PM.
you'll need to connect stuff that originally connects to `audioContext.destination` to a main gain node that **is** connected to `audioContext.destination`, you already make gain nodes for playing notes so you can just make a main gain node here

you can then register the context with this function:
```js
runtime.registerExtensionAudioContext("EXTENSION_ID", audioContext, gainNode);
```
make sure to replace EXTENSION_ID with your extension's ID.

this will make the pause button and volume slider work with the extension, and future addons or extensions will be able to modify it if needed

JeremyGamer13:
you should rename your files so they dont have spaces in the name

JeremyGamer13 (changes requested):

PR #261, opened 2024-12-09
Soup Vectors Extension

the-can-of-soup:
This is an extension I made that adds 36 vector math-related blocks. Extension ID: "souputilsvectors"

Name: Soup Vectors v4
Author (PenguinMod username): Soup
Description: Vector math tools so you'll never need to compute an arctangent ever again.

Thumbnail:
soup_vectors_thumbnail


AlexSchoolOH:
You might want to take the Campbells out of the thumbnail, its a copyright concern for PM

the-can-of-soup:
OK I'll work on that

the-can-of-soup:
I removed the "Campbell's" text from the logo

the-can-of-soup:
I just found a Unicode error in the code, so don't merge this until I fix it

CoolDevXD:
i made something like this, yours is probably way better tho.

the-can-of-soup:
1d, 2d, 3d, 4d, whatever you want

derpygamer2142:
1d is just a variable

the-can-of-soup:
yeah ik but you can do it if you want for some reason

the-can-of-soup:
that will be an upcoming feature but for now you can use the json extension "in array (... your vector) get (3)" for grabbing w and you can use "in array (vector (x), (y), (z)) add (w)" to construct a vector

the-can-of-soup:
im gonna start working on v4.4 which will add the 4d blocks

the-can-of-soup:
ok all the 4d stuff is there now

PR #263, opened 2024-12-14
update penguingpt.js so it uses the default proxy

AdamMady:
set the default api url block value to latest default proxy set by mubi instead of setting it to the old and shut down "https://proxy.mubilop.tech/v1"


derpygamer2142:
remember to mirror this pull request with https://github.com/PenguinAI-Ext/extension/pulls so we have the same version of the extension on each repo

MiloDev123:
@AdamMady change the version to v1.95, i do not have enough permissions to do that here.

AdamMady:
> @AdamMady change the version to v1.95, i do not have enough permissions to do that here.

alright

MiloDev123:
@AdamMady Newest Proxy URL is `https://api.penguinai.tech/v1`, please change that.

MiloDev123:
This is still not merged. This version is already old, i will make a new pull request.

MiloDev123:
Newest Pull Request is #267

PR #267, opened 2025-01-22
Update penguingpt.js

MiloDev123:
Newest version

### DIFF
--- a/static/extensions/MubiLop/penguingpt.js
+++ b/static/extensions/MubiLop/penguingpt.js
@@ -8,7 +8,7 @@
 //                 |___/                                        
 //
 // == Harness the power of AI in your projects! ==
-// (YOU DO NOT NEED TO LOAD UNSANDBOXED)
+// (YOU NEED TO LOAD UNSANDBOXED)
 // By LOLEMO, Forked by Anonymous_cat1 and then forked by MubiLop
 // Only tested on Penguinmod.com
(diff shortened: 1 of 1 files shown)


static/extensions/MubiLop/penguingpt.js
```diff
@@ -8,25 +8,25 @@
...
+if (!Scratch.extensions.unsandboxed) {
+    throw new Error('"PenguinAI" cannot run unsandboxed.');
+}
+
+let api_url = 'https://api.penguinai.tech/v1';
 
-    // Initialize the api_url variable with a default value
-    let api_url;
-    fetch('https://mubilop.tech/proxy')
+fetch('https://mubilop.tech/proxy')
```
JeremyGamer13, line 29: preferably dont request anything right when the extension is imported
if you **need to do a request** here, use Scratch.fetch

MiloDev123: > preferably dont request anything right when the extension is imported if you **need to do a request** here, use Scratch.fetch

That creates an Extension Security popup which might be annoying for the user / might not make the user comfortable with using the extension.

derpygamer2142: that's the point

RedMan13: yea thats the point, all requests are supposed to be routed through the popup

PR #274, opened 2025-02-28
Change my extension to Twitch -> Streamer, also huge update?? ig (PREFORMANCE, OPTIMIZATION)

liablelua:
twitch was against scratch & twitch tos so this one isn't lol


AlexSchoolOH:
Changing the folder and filename is not advisable because any projects that use the extension will break due to a dead link.

SharkPool-SP:
you shouldnt be removing extensions. As that will break old projects that used them. You should just hide it. Once you fix that, ill review this extension

PR #275, opened 2025-03-01
Remove dead captcha extention

1TSnakers:
I found out that the captcha extension is unusable as the API (captcha.penguinmod.com) doesn't work, I just removed it here, you can either keep this, switch it out with shark pools, or fix the API

### DIFF
--- a/src/lib/extensions.js
+++ b/src/lib/extensions.js
@@ -251,22 +251,22 @@ export default [
         creator: "SharkPool-SP",
         isGitHub: true,
     },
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
@enderhacker this is your extension, fix it at some point

enderhacker:
Will fix asap
Eta 2 days max
@SharkPool-SP

1TSnakers:
@enderhacker Might I suggest a block to determine if the server is down?

PR #281, opened 2025-03-16
Add Enumerations Extension

DogeisCut:
I've created yet another extension. This one adds Enumerations seen in many programming languages to the editor!
msedge_VtFqoEGerF
msedge_Tk4Vsk1hTB

They are properly serialized and will save+load correctly. The dropdown of the first block also updates to the enum you have selected.

This is the icon
Enumerations


static/extensions/DogeisCut/Enumerations.js
```diff
@@ -0,0 +1,387 @@
+// Name: Enumerations
+// ID: dogeiscutenumerations
+// Description: No description provided.
+// By: DogeisCut <https://scratch.mit.edu/users/DogeisCut/>
+// License: MIT
+
+// Version V.1.0.0
+
+// TODO:
+// - Use dynamic blocks instead of this opcode fuckery
```
JeremyGamer13, line 10: imo i think it would be best to do this before this is added to the library, the current method feels unsafe & switching to dynamic blocks later would require breaking changes to the extension

JeremyGamer13 (changes requested):
(see other messages in review aswell)

in my opinion i think a better description of what the point of enumerations are would be useful, would be good for the description of the ext

static/extensions/DogeisCut/Enumerations.js
DogeisCut: Will do

PR #282, opened 2025-03-16
Add FormatNumbers Extension

DogeisCut:
This is a simple extension that lets you format large numbers into AD Standard, Fixed Decimal, Comma Separated, or Scientific Notation.

It adds but one block, yet a very useful one imo, especially for incremental games!

msedge_2ZeCztrapG
msedge_b8scWM1ITL
msedge_easzgrLqwB
msedge_bvo6IzMFhC
msedge_d6Rc1Kz7ia
msedge_i5PasinNgF
msedge_SZVAPgfXeq

This is the banner, drawn by Dillon (dillonr on Discord)!
FormatNumbers


static/extensions/DogeisCut/FormatNumbers.js
```diff
@@ -0,0 +1,216 @@
...
+                type: Scratch.ArgumentType.NUMBER,
+                defaultValue: 2,
+              },
+            },
+          },
+        ],
+        menus: {
+          notationOptions: {
+            acceptReporters: false,
+            items: [
```
JeremyGamer13, line 203: Scratch.translate is used earlier but these options are not translatable, is that intentional?

JeremyGamer13 (changes requested):
just a note, the extension returns an error if the decimal place number is less than 0 or greater than 100, but it might actually be good to keep it as an error. thats up to you though
image

static/extensions/DogeisCut/FormatNumbers.js
DogeisCut: Was just cause I was formatting this for Turbowarp as well, this is intentional. I may change that error however.

DogeisCut:
I mistyped "translated" as "translatable" if you're wondering what that force push was about, but then i realised putting translatable made more sense anyways

static/extensions/DogeisCut/FormatNumbers.js
DogeisCut: Uh, no I actually forgot to make the dropdowns translatable, will be fixing that as well as the error since I'm keeping this compatible with Turbowarp.

DogeisCut:
as for the description i had a bit of a dummy moment and mis interpreted what you said and forgot to change it

PR #291, opened 2025-04-09
Extension Fix: AllMenus Missing Menus

DogeisCut:
Adds the costumes, backdrops, and keys menus. They were missing before.


yuri-kiss:
was this fixed in the TW version

Steve0Greatness:
After glancing at TurboWarp's version, it doesn't appear that it was. This work should probably be sent up to them as well.

PR #292, opened 2025-04-09
Upload new extension banner

Ikelene:
changed banner so it fits in more with the other extensions.


Ikelene:
ExtensionBanner
banner in question

PR #301, opened 2025-04-26
Add extension: Extra Timers 

Steve0Greatness:
Supersedes #299. That one was closed due an automated action by GitHub that closes PRs when the branch they're meant to pull from is deleted or renamed.

This extension allows for the creation and management of timers, similar to the regular timer under sensing.

Blocks currently in this extension:

The blocks appearing in this extension

Here is the extension thumbnail:
3 digital clocks in red, green and then blue at varying x locations placed uniformly along the y axis.


Steve0Greatness:
I feel like this I've fixed pretty much anything I can currently. If there's any major bugs, please let me know.

Steve0Greatness:
FYI, this is the current thumbnail I have

image

JeremyGamer13 (changes requested):
- Pausing a timer while its already paused or stopped will increase the time (not sure if this is a feature or a bug)
- Upon saving & reloading a project, then updating the toolbox (switching to costume editor or switching sprites), strange things can happen with sprite-only timers. The code editor also crashes if you load the project twice(?) [images]
- "Timers for this sprites"
- > Note: the change I just pushed breaks compatibility with custom styles. (Or at least does hypothetically)
can confirm it does, you might want to add a warning on the extension page for it (you cant set value background color)

image
image
image

Steve0Greatness:
> Upon saving & reloading a project, then updating the toolbox (switching to costume editor or switching sprites), strange things can happen with sprite-only timers. The code editor also crashes if you load the project twice(?) [images]

First one, I can't replicate.

That's... odd. Seems to be that somehow the "name" property never got initialized. I'll have to look a little more into that, I'm not sure how it wouldn't be initialized.

Steve0Greatness:
I think I fixed the crash by just copying more from the [XML escape](https://github.com/PenguinMod/PenguinMod-Vm/blob/develop/src/util/xml-escape.js) utility.

Steve0Greatness:
Still haven't replicated the strange behavior with for this sprite timers. I think I fixed both of them with that last push... I think.

Steve0Greatness:
> you might want to add a warning on the extension page for it (you cant set value background color)

Is there a specific JSON key for marking incompatibility? Incompatibility doesn't mean it's really unstable IMO.

Steve0Greatness:
Shouldn't be incompatible anymore.

Steve0Greatness:
Only built-in blocks are able to set themselves as per-sprite, so every time any block in this extension is run anywhere, it always assumes that the variables are global, so `util.target` returns the editing target...

Steve0Greatness:
> * Upon saving & reloading a project, then updating the toolbox (switching to costume editor or switching sprites), strange things can happen with sprite-only timers

Weirdly enough, I'm not sure this is a fault in my extension.

Steve0Greatness:
I haven't seen anything being switched to being local, but I have seen timers being duplicated, literally duplicated but with an ID of undefined.

JeremyGamer13:
> Incompatibility doesn't mean it's really unstable IMO.
It's really just an extension warning label, it's named `unstable` since that's the most common reason for it to be on an extension

Steve0Greatness:
> It's really just an extension warning label, it's named `unstable` since that's the most common reason for it to be on an extension

Either way, it's not incompatible anymore.

JeremyGamer13:
image
Not sure what's causing this anymore but reloading the toolbox makes timers called "undefined" appear. They get renamed to "null" after saving again and reloading

Steve0Greatness:
I think... I fixed it... maybe

JeremyGamer13 (approved):

PR #311, opened 2025-05-21
Add Email2FA Extension

Ikelene:
extension that allows users to send a verification code to an email and if the user enters the correct email they get verified.
might change name or something later but for now its fine
also might add blocks to verify per project instead of globally but thats for later


Ikelene:
why did it fail
one sec ill figure it out

Ikelene:
nvm i cant even look at the vercel report pls tell me why it failed then

JeremyGamer13:
> nvm i cant even look at the vercel report pls tell me why it failed then

syntax error after creatorAlias

Ikelene:
all fixed

Ikelene:
might change banner though it looks kinda ass 💔

Ikelene:
added per project verification

Ianyourgod:
i see a few problems. i’ll write a review soon

Ikelene:
> i see a few problems. i’ll write a review soon

Okay thanks

Steve0Greatness:
Not a requirement, but just kinda curious. Are you intending on releasing the source code of the backend, or is it going to remain closed source for the time being?

Ikelene:
> Not a requirement, but just kinda curious. Are you intending on releasing the source code of the backend, or is it going to remain closed source for the time being?

Closed for the time being for security concerns. it may be released later.

Steve0Greatness:
> might change banner though it looks kinda ass 💔

New version looks a little better, but I still have a few issues with it.

2FA_ExtensionBanner

Steve0Greatness:
> Closed for the time being for security concerns. it may be released later.

Alright, in the case that the source code is released, could the extension also have a feature to allow for the server to be changed?

Ikelene:
> Alright, in the case that the source code is released, could the extension also have a feature to allow for the server to be changed?

Yeah, but I dont see how that would really change much. All it would do is change the email it comes from and possibly the formatting which may come as support later or different variants.
How would this feature benefit you?

Steve0Greatness:
> Yeah, but I dont see how that would really change much. All it would do is change the email it comes from and possibly the formatting which may come as support later or different variants.
> How would this feature benefit you?

The primary idea is that an extension shouldn't rely exclusively on a single webservice when possible. Additionally, some people may want either of those, such as if they're hosting their project on their own website, or may want the style of the email to match their project.

Ikelene:
> i see a few problems. i’ll write a review soon

wheres my review ian

Ianyourgod:
erm
somewhere

Steve0Greatness:
The path listed in `src/lib/extensions.js` is incorrect. It should actually be `Ikelene/email2FA.js`, remember that paths are treated as case-sensitive.

static/extensions/Ikelene/email2FA.js
```diff
@@ -0,0 +1,196 @@
+(function(Scratch) {
+    'use strict';
+
+    let cachedProjectID = null;
+
+    class TwoFAExtension {
+        getInfo() {
+            return {
+                id: 'twofa',
```
Ianyourgod, line 9: I would suggest changing this to like "ikelene_twofa" since "twofa" seems like something that could be 
accidentally used by another extension creator pretty easily

static/extensions/Ikelene/email2FA.js
```diff
@@ -0,0 +1,196 @@
...
+                name: 'Email 2FA',
+                color1: '#ff8c3b',
+                color2: '#ff8c3b',
+                blockIconURI: 'https://ikelene.ca/api/2fa/icon-white.svg',
+                menuIconURI: 'https://ikelene.ca/api/2fa/icon-orange.svg',
+                blocks: [
+                    {
+                        opcode: 'registerProjectID',
```
Ianyourgod, line 17: maybe instead of registering a name, have the block just be "register project", generate a unique id, and store it in the project. then you don't have to worry about naming collisions

Steve0Greatness:
What is your rate limit set to? I swear it seems like it was reached in a single usage of the block. That makes it incredibly difficult to test at all.

Also, is `cachedProjectID` supposed to be saved across loads, because to do that you'd need to add  `serialize` and `deserialize` functions.

Ianyourgod (commented):
small little issues, overall it seems good

static/extensions/Ikelene/email2FA.js
Steve0Greatness: That's actually a kinda good idea. That could probably even be done on load like the Storage extension does.

Steve0Greatness:
> Closed for the time being for security concerns. it may be released later.

What kind of security concerns? API keys? Arbitrary code execution? Even in the case that those do exist, the code being open source would generally be better, since it'd mean that others would be able to see and thus potentially fix such security issues.

Steve0Greatness:
> its just a pm extension banner its not that deep gng 🥀

It kinda is, though. Visual design is incredibly important.

static/extensions/Ikelene/email2FA.js
Ikelene: I name it because thats what shows up in the email.
`Your verification code from {projectName}`

Ikelene:
> What is your rate limit set to? I swear it seems like it was reached in a single usage of the block. That makes it incredibly difficult to test at all.

You can send an email every 60 seconds

static/extensions/Ikelene/email2FA.js
Ianyourgod: then store the name too. its still better to have a unique id instead of relying on the user for that.

Ikelene:
Why would the name need to be stored? It's already on the server when you get your project id to use in the emails

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

Ianyourgod:
i meant like store it server side

Ikelene:
It is stored server side. The server keeps all the project ID's and names.

Ikelene:
Before you push this into the main repo, I will be making the following changes and additions:

- A block to let you pick from a few presets of how the emails will look.
- A section to let you customize the email with custom html
- A section to use your own email address if you dont want to use the default.
- The custom HTML feature will only work if you use your own email, if you use the default you have to pick from the presets.

Steve0Greatness:
> It is stored server side. The server keeps all the project ID's and names.

Not sure if you noticed, but your commit that is supposed to add the project ID to the project files doesn't actually do that. Again, you need a serializer and deserializer.

Ikelene:
changes have been applied it is ready to be pushed now

Steve0Greatness:
Would you mind explaining to me what `vm.extensionState` is? I've never heard of it.

Steve0Greatness:
Also, the `Scratch` object doesn't provide access to the `runtime` object, to get that you need to put in `Scratch.vm.runtime`.

Steve0Greatness:
Also, once more, could the extension actually use the cached ID rather than requiring every block to have it passed to it?

static/extensions/Ikelene/email2FA.js
```diff
@@ -0,0 +1,317 @@
...
+ { text: 'Minimalist Dashed', value: '3' },
+ { text: 'Welcome Style', value: '4' }
+                        ]
+                    }
+                }
+            };
+        }
+
+        constructor() {
+            const runtime = Scratch.runtime;
```
Steve0Greatness, line 164: The runtime is literally passed to the constructor as the first argument. Save the runtime under the current object rather than grabbing a reference to the runtime every time. Like,
`this.runtime = runtime`.

static/extensions/Ikelene/email2FA.js
```diff
@@ -0,0 +1,317 @@
...
+            if (runtime && runtime.vm) {
+                const state = runtime.vm.extensionState['IkeEmailTwofa'] ||= {};
```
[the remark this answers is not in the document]
Ikelene: i have no idea how to do the saving and chatgpt is no help either, it uses anything but serialize and deserialize

static/extensions/Ikelene/email2FA.js
```diff
@@ -0,0 +1,317 @@
...
+
+        setEmailPreset(args) {
+            emailPreset = parseInt(args.PRESET);
+        }
+
+        setCustomHTML(args) {
+            customHTML = args.HTML;
+        }
+
+        setCustomEmailLogin(args) {
```
Steve0Greatness, line 181: Why are you encouraging people to put in their email passwords?

Steve0Greatness: If they're distributing their project, it's quite literally just going to be them distributing their email password in plaintext. If you're storing it on your servers, you'd also need to store it in plaintext, or a facsimile of plaintext that does nothing, like encrypting the password with a commonly accessible encryption key. This is genuinely awful for security and I'm extremely shocked that you didn't think about this.

Gen1xLol: > If they're distributing their project, it's quite literally just going to be them distributing their email password in plaintext. If you're storing it on your servers, you'd also need to store it in plaintext, or a facsimile of plaintext that does nothing, like encrypting the password with a commonly accessible encryption key. This is genuinely awful for security and I'm extremely shocked that you didn't think about this.

what he said

static/extensions/Ikelene/email2FA.js
```diff
@@ -0,0 +1,317 @@
...
+                    //{
+                    //    opcode: 'clearCustomEmailLogin',
+                    //    blockType: Scratch.BlockType.COMMAND,
+                    //    text: 'Clear custom email login'
+                    //}
+                ],
+                menus: {
+                    presetMenu: {
+                        acceptReporters: true,
+                        items: [
```
Steve0Greatness, line 151: Could this be generated by the server. Like having the contents of this menu dynamically gotten from the server. That could be done by setting this (`items`) to a string with the same name as the function that returns an array for this list. This could allow for the dynamic adding of templates to your server without needing to update your extension.

static/extensions/Ikelene/email2FA.js
```diff
@@ -0,0 +1,317 @@
...
+
+        async verify2FACode(args) {
+            try {
+                const res = await fetch('https://ikelene.ca/api/2fa/verify_code.php', {
+                    method: 'POST',
+                    headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+                    body: new URLSearchParams({
+                        email: args.EMAIL,
+                        code: args.CODE,
+                        project_id: args.PROJECT_ID
```
Steve0Greatness, line 268: For, like, the third time, could this be gotten from the project ID stored in the cache rather than being taken in as an argument?

static/extensions/Ikelene/email2FA.js
```diff
@@ -0,0 +1,317 @@
...
+                });
+
+                const id = await res.text();
+                const trimmed = id.trim();
+
+                const isError = ['invalid name', 'invalid chars', 'rate limit', 'rate limited'].includes(trimmed);
+                if (trimmed && !isError) {
+                    cachedProjectID = trimmed;
+
+                    const runtime = Scratch.runtime;
```
Steve0Greatness, line 211: Store your runtime directly in your extension by adding it on construction rather than getting it from the sandbox every single time. Also, I think I've already said that it's passed into the constructor as the first argument.

static/extensions/Ikelene/email2FA.js
Steve0Greatness: > i did think about it last night and i will be removing it and revamping how that would work, or possibly just open source the backend

I think the second is better, since I can't think of anything else that wouldn't require a lot of undue trust given to any given party.

static/extensions/Ikelene/email2FA.js
Steve0Greatness: This would be especially nice if you ever decide to open source the backend, since it would allow for people to fork your backend to add more email templates without needing to create a fork of the extension (which is essentially the whole goal of allowing for the user to change the server).

static/extensions/Ikelene/email2FA.js
Steve0Greatness: Pretty much just replace any `args.PROJECT_ID` with `cachedProjectID` in your codebase, and remove all `PROJECT_ID` arguments from the `getInfo` method.

static/extensions/Ikelene/email2FA.js
Steve0Greatness:
```js
constructor(runtime) {
  this.runtime = runtime;
}
```

Steve0Greatness: Actually, nevermind.

Steve0Greatness:
I sent a PR to your repo that allows for the extension to properly store values between project loads.

TheShovel:
i would recommend not merging this until the backend is open source due to privacy concerns. This would have to be handled in the PM privacy policy somehow if any data at all is stored

Steve0Greatness:
> why does any of this matter? The extension is not being provided by the server, so i dont see why it needs a similar license or why we need to make changes to our privacy policy. I think the service would need the privacy policy, no? We have other extensions that fetch APIs too.

A similar license wouldn't be necessary, it's basically just for consistency. As for the privacy policy thing, I believe it's because this API would be specific to the extension, rather than being more general like Discord or Google's OAuth.

Steve0Greatness:
> why tf does ts need a license 🥹🥹
> ill open source backend soon

The license is the most important part of open sourcing. If it isn't an open source license, it isn't open source.

Gen1xLol:
what a PR

PR #314, opened 2025-06-04
Add Content Filter

Ikelene:
Extension that contacts my advanced API for filtering words. It even filters context, so something like `You should kill yourself` becomes `You should *************`

### DIFF
--- a/src/lib/extensions.js
+++ b/src/lib/extensions.js
@@ -184,7 +184,7 @@ export default [
     {
         name: "Content Guard",
         description: "Includes utilities related to filtering bad words, swearing and profanity.",
-        code: "Ikelene/content_gaurd.js",
+        code: "Ikelene/content_guard.js",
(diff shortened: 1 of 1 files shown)


Preloading:
Why does this need to contact a server? 

In DMs you stated it's because you are using AI, but should AI really be used in this scenario? In my testing, for long strings it breaks it very badly (only censoring parts of it), and if the request fails for any reason, it will pass the unfiltered text straight to the user, with no way of telling it failed. It also can take sometimes 18 seconds for a single request to complete.

TheShovel:
this extension is INCREDIBLY unreliable. It can return strings of text that are not even related to what you said due to how it works. Ai si good for text filtering, but not like this... there are better implementations out there that dont rely on the ai returning the exact text you need. Also its incredibly slow at times

I will close this pr because i dont think this is fixable without a whole rewrite of how the api it uses works

PR #322, opened 2025-06-20
PenguinAI: Small updates

ScratchFakemon:
Update some of the prompts to have better grammar, add a second line of text to the model selector that says PenguinAI might be down if it takes too long, and maybe something else i forgor just check the diff lol


cicerorph:
> Update some of the prompts to have better grammar, add a second line of text to the model selector that says PenguinAI might be down if it takes too long, and maybe something else i forgor just check the diff lol

Please, do this pull request to https://github.com/PenguinAI-Ext/extension other than the actual version on PenguinMod, we will update the prompts on the next update batch pull request.

PR #327, opened 2025-06-25
Create Performance.js

DillonRGaming:


Steve0Greatness:
You should add this into `src/lib/extensions.js`, and also add your image under `static/images/DillonRGaming/`.

PR #328, opened 2025-06-28
Create HowWeFeel-extension.js

Nedas851:
Can you add my very first extension named **"How We Feel+ (Real-time)"**

(too lazy to create thumbnail of extension so you have to bear with me with this)


FreshPenguin112:
We don't accept turbobuilder/extforge extensions, sorry

jwklong:
it clearly isn't a extforge-made extension, i dont know why it says so in the comments but that doesn't even have the special `ExtForge` constant

SharkPool-SP:
This is not even remotely accurate/useful since youre detecting 'emotion' through the cameras brightness. 💀

TheShovel:
this is probably ai generated. Also it doesnt do what it says it should do

PR #336, opened 2025-07-08
Updated googleAuth.js to work with the new API

Ikelene:
This should be merged as soon as possible as all current projects using the extension are incompatible with the new API and will not work.


Ikelene:
Wait, dont merge just yet I might have found a bug

Ikelene:
all fixed ready for merge

cicerorph:
(btw, dont make updates that break the old extensions :D, you should prob soft deprecate, not just remove features entirely, only if its a api url)

PR #338, opened 2025-07-15
New Windows extension

Monochromasity:
Lets you open websites in new windows.

I didn't know how I was meant to provide an extension image so I just uploaded it into the Monochromasity folder in the images folder (it is also at the bottom of this PR description). I spent a long time making this so it would be nice if it was added

Let me know if there's anything you want me to change, I'm happy to do so :)

newwindows


Monochromasity:
> its ridiculous how long that block is. I dont want to merge this like that because it would be really hard to work with in a project. Please split it into multiple blocks, or blocks with drop downs to set properties

I didn't think it would be an issue but considering the fact that some of the properties aren't as important, I'll make a "set property" block for individual properties, and only keep the width, height, distance from left, and distance from right in the actual open window block to save block usage in practical use.

PR #339, opened 2025-07-21
Create grannyscookies.js

DashDevmationsDash:
Its My Extension!

Name:Granny's Cookies

Thumbnail:
costume1

Description:
An Extension That Adds Th Abillity To Create And Read Cookies To Save User Info


DashDevmationsDash:
> Extforge extensions/Extensions made with builders aren't allowed

And Why?

PR #340, opened 2025-07-22
Addons extension

PuzzlingGGG:
make sandboxed js addons that interact with the project, only where you give it access to tho. can be used for like modding apis and such. kiiindaaaa cooool


Steve0Greatness:
> Isn't this just the JavaScript extension but with more blocks

This looks like it's intended more towards being able to allow for plugins or mods to be created.

PR #341, opened 2025-07-23
Extension update: Add 12 new blocks to 3DMath.js

LSPECTRONIZTAR:
Screenshot_20250723_131558_Chrome
Adds 12 new blocks to `3DMath.js` that might be useful.

I coded this on my phone


TheShovel:
some of the blocks return NaN. We shouldnt have that. They should default to a value like 0 or something. Unless there is a reason for it, i cant merge in the current state

TheShovel:
> > some of the blocks return NaN. We shouldnt have that. They should default to a value like 0 or something. Unless there is a reason for it, i cant merge in the current state
> 
> Could you please explain which blocks and why? If it's the angle vector it will return NaN if you use a zero vector like [0,0,0]. I tried to fix it, but that's just how it is.

yeah i was talking about the angle vector. Is this at least consistent with other blocks like that? If not, make it return [0,0,0]

LSPECTRONIZTAR:
> > > some of the blocks return NaN. We shouldnt have that. They should default to a value like 0 or something. Unless there is a reason for it, i cant merge in the current state
> > 
> > 
> > Could you please explain which blocks and why? If it's the angle vector it will return NaN if you use a zero vector like [0,0,0]. I tried to fix it, but that's just how it is.
> 
> yeah i was talking about the angle vector. Is this at least consistent with other blocks like that? If not, make it return [0,0,0]

Ok thanks, I'll fix it
Also I'm going to combine the two reflect blocks into one because it makes more sense (no one has probably tried to use the updated extension in a project anyway lol)

LSPECTRONIZTAR:
Alright I fixed the angle block and also merged both reflect blocks but hid the old one for compatibility just in case

LSPECTRONIZTAR:
Thank you!

PR #342, opened 2025-07-26
Extension Submission - "MIDI Devices"

jexapatron:


static/extensions/jhub7682/MIDIdevices.js
```diff
@@ -0,0 +1,182 @@
...
+            break;
+
+          case 0xA0: // Polyphonic Aftertouch
+            polyAftertouchValues.set(data1, data2);
+            break;
+        }
+      };
+    }
+
+    Scratch.extensions.register({
```
JeremyGamer13, line 66: you really shouldnt register your extension like this. make a class and put your extension's functions in there, then register an instance of that class
https://docs.penguinmod.com/development/extensions/starting-out#lets-start-simple

JeremyGamer13, line 66: the extension should always register even before you ask for MIDI permission, if you dont do this then people's projects will fail to load properly since the extension wont be loaded before the project loads

static/extensions/jhub7682/MIDIdevices.js
```diff
@@ -0,0 +1,182 @@
...
+      getInfo() {
+        return {
+          id: "MDjhub7682",
```
JeremyGamer13, line 69: i would make this ID more clear with something like `jhub7682MidiDevices`
just so if someone gets an error with just the extension ID it makes more sense what the extension is actually called

JeremyGamer13 (changes requested):
small review before i actually test this with a real midi device

PR #343, opened 2025-07-27
Waiting Family

Gloxk12sUltra:
costume3(1)

Adds More Waiting Blocks.


Gloxk12sUltra:
my user is Microwavedo on penguinmod

also this adds more waiting blocks instead of just wait (num) seconds block

Gloxk12sUltra:
> Not only is the extension easily re-createable with existing blocks, this is made with extension forge. Which we don't accept here

so which thing should i use to create extensions instead of forge?

PR #347, opened 2025-07-30
Sparrow

SkyBuilder1717:
Sparrow V2 is a format for exporting sprite sheets to Adobe Animate. The export creates PNG and XML format. PNG contains all the frames that were in the selected symbols. XML contains data about frames to be used later in Flixel for sprites.
browser_wZsEXDwFiG


TheShovel:
we already have an extension that supports sprite sheets. I dont think this one does anything special compared to the existing one. If there is a feature missing, in the existing one, please just make a PR to modify the extension that already exists

JeremyGamer13:
> I dont think this one does anything special compared to the existing one.

The main appeal seems to be that this extension parses a format that Adobe Animate can export. I think Sparrow v2 doesn't include duplicate frames & can rotate things to make them fit (unless that's another format I'm thinking of.)

TheShovel:
This should still just be included in the current extension

PR #361, opened 2025-08-14
Bitwise+ v1.3

BurningHot687:
This is an expansion of the Bitwise extension on Turbowarp with some more capabilities.


BurningHot687:
Also last day of summer so this special project of mine (first extension :0) won't take up too much school time.

BurningHot687:
Alright I don't know if I have to delete the png because it'll choose the png instead of the avif or if avif is supported yet, so can anyone tell me?

PR #367, opened 2025-09-04
SuperCodes_BetterKeyboard

moosanaeempc-lgtm:


AlexSchoolOH (changes requested):
Here are some suggestions

Issue #365, opened 2025-08-28
MoreFields & Extension exposer - Licenses do not follow the guidelines


yuri-kiss:
Both of these are my fault entirely. Fix is to update the guidelines, otherwise MoreFields violates the guidelines either way, ~~Extension exposer can just be fixed by reverting to before the violating commit OR updating the guidelines.~~
According to `Your extension should be MIT licensed (not compatible with the MIT license, should be the actual MIT license. This may change in the future.)` the files
- `static/extensions/Ashime/MoreFields.js` (violating commit: https://github.com/PenguinMod/PenguinMod-ExtensionsGallery/commit/690e8a1a024ea04aea37c8471a6e34c10a8dda47)
- `static/extensions/TheShovel/extexp.js` (violating commit: https://github.com/PenguinMod/PenguinMod-ExtensionsGallery/commit/22df1812be097db5635159972910b2e23d5d31f7)

don't follow the requirements because of the `LGPL-3.0`/`LGPLv3` license attachment.
If I am mistaken just close this issue, and possible update the README to make this more clear?

Currently if https://github.com/PenguinMod/PenguinMod-ExtensionsGallery/pull/419 gets merged then this issue will be solved correctly without rewriting anything (extension exposers license will not be changed back, and will stay as MIT).

JeremyGamer13:
the best option for More Fields is likely just to rewrite it under MIT, I will try to do this soon
this is similar to how TW currently handled Animated Text

for extension exposer i'll see if reverting removes anything

JeremyGamer13:
the licensing rule is basically just so everything is 100% consistent, everything uses the same terms so if you can use 1 extension you can use all of them

Steve0Greatness:
On second thought, it might be fine to just be reverted.

JeremyGamer13:
> whatever code was contributed

basically the whole thing was reformatted so it would basically have to be reverted anyway

RedMan13:
idk, what is lgpl

Steve0Greatness:
> what is the whole issue in allowing lgpl, it just means others have to copy the license if they modify the extension so it doesnt affect the end user

Jeremy wants everything to have a consistent license so that the rules regarding it are always guaranteed to be the same.

Steve0Greatness:
> its not inputs though, its fields meant to demonstrate the built in scratch api... youd be ruining the point of the extension

[...]

PR #374, opened 2025-09-07
Reimplement current Extension Exposer in old version

JeremyGamer13:
Reimplements the current functionality of Extension Exposer into the old version. Essentially a revert but hopefully without breaking compatibility.
The purpose of this Pull Request is to make Extension Exposer's licensing consistent with the rest of the repositories' licensing (MIT-only).

Related:
#365


JeremyGamer13:
I understand on removing your name, I can do that.

> you removed functions and changed how the blocks function

I believe the blocks should do the same things, I made a quick test project that worked in both versions but I am also trying to look for projects which use the extension to see if they break (that's why this isn't merged right now)

JeremyGamer13:
Superceded by #375

PR #375, opened 2025-09-09
Simplify and relicense extension exposer.

yuri-kiss:
~~For licensing reasons we need @TheShovel to approve of this change.~~ Done

The simplification just simplifies the block adding process (push is faster than constantly checking the length)

This also fixes a goofy bug that happens when an extension implements both of the APIs (scratch and standard)

Super scedes: #374 and fixes license issue 1 in #365


JeremyGamer13 (approved):

static/extensions/TheShovel/extexp.js
```diff
@@ -89,24 +89,25 @@
...
+      const ext = runtime[`ext_${EXTLIST}`];
+      const blocks = [];
+      // Check if the extension implements the standard extension API
+      if (ext && (typeof ext.getInfo === 'function')) {
+        const info = ext.getInfo().blocks;
+        if (!info) return blocks;
         for (let index = 0; index < info.length; index++) {
-          const element = info[index];
-          addonBlocks[addonBlocks.length] = element.opcode
+          blocks.push(info[index]);
```
JeremyGamer13, line 100: forgot to check this, this makes the list show [object Object] for each block

PR #381, opened 2025-10-02
Extension: Objects

DogeisCut:
yes yes i know there are already json extension. But this one in specific works with jw's array extension and uses several penguinmod specific APIs 

dogeiscutObject
image

@jwklong said they wanted this in the main/front page gallery so do with that information as you will


DogeisCut:
yeah so the thumbnail is already the correct size for main gallery so you dont have to worry about that

Ianyourgod:
how many json extensions do we need bruh :sob:

RedMan13:
olive them :trollface:

LordCat0:
shlawg we have enough json

DogeisCut:
> how many json extensions do we need bruh 😭

idk ask @jwklong

PR #385, opened 2025-10-03
Objects: fix another [object Object] issue

DogeisCut:
last one i swear


DogeisCut:
wait no i was looking at the wrong version... wow im good at this

PR #387, opened 2025-10-03
Objects: Fix null/undefined not getting parsed from JSON

DogeisCut:
yay
image


DogeisCut:
ok it saves fine somehow... what does it save fine is arrays! @jwklong might want to fix this
image

PR #419, opened 2025-11-21
Allow LGPL-3.0 licensing

JeremyGamer13:
Making this a draft since I'm not sure what is required of submitters or us to allow this.
I just updated the README.md but maybe more needs to be said/added to the repo 🤷 
If someone knows then that would be nice

### DIFF
--- a/static/extensions/LICENSING.md
+++ b/static/extensions/LICENSING.md
@@ -5,10 +5,20 @@ PenguinMod extension code in this folder is under one of two licenses:
 
 For an extension to use a specific license, it should specify the [SPDX short identifier](https://spdx.org/licenses/) (seen above) in a comment at or near the top of the file.
 
-Licensing example:
+Licensing examples:
 ```js
 // By: PenguinMod <https://github.com/PenguinMod>
 // License: LGPL-3.0
+(function (Scratch) {
+    // ...
+```
+```js
+/**!
+ * @file My Very Cool Extension
+ * @author PenguinMod
+ * @link https://github.com/PenguinMod
+ * @copyright Licensed under LGPL-3.0
+ */
 (function (Scratch) {
     // ...
 ```


README.md
```diff
@@ -16,9 +16,9 @@ Create your extension. You can find a lot of details [here](https://docs.turbowa
 and [Gallery page](https://extensions.penguinmod.com/) to confirm.
 - Your extension should be created for a specific purpose. An extension with random blocks someone might need will likely not get accepted if the blocks are not in a certain theme, like math or rendering.
 - Your extension should be easily explainable in 1-2 sentences or a single picture. This is what will determine whether you can be on the gallery or not.
-- Your extension should be MIT licensed (not compatible with the MIT license, should be the actual MIT license. This may change in the future.)
-    - `Ashime/MoreFields` is currently a **temporary** exception to this rule. (MIT & LGPL-3.0 license)
-    - The above exception is an accidental change. Do not knowingly submit extensions under a seperate license to purposely cause licensing issues on our repository.
+- Your extension should be under the MIT license or the LGPL-3.0 license.
+    - You cannot make or use a different license that is compatible with one of the above licenses. You must use one of the specified licenses exactly.
+    - Do not knowingly submit extensions under a separate license to purposely cause licensing issues on our repository.
```
Steve0Greatness: This line, 21, feels redundant due to the previous line.

yuri-kiss:
> I think this is pretty much all that needs to be said to allow this, given this is the only part of the repo that specified a license before. Might want to consider allowing more permissive and weak copyleft licenses at some point in the future, though, in case something like what happened prior happens again.

like MPL?

JeremyGamer13:
@Steve0Greatness @yuri-kiss 
this may be good? only thing left is i am not sure how dual-licensing with MIT and LGPL plays into this (see https://github.com/PenguinMod/PenguinMod-ExtensionsGallery/blob/main/static/extensions/Ashime/MoreFields.js#L5C4-L5C35 )

yuri-kiss:
> @Steve0Greatness @yuri-kiss this may be good? only thing left is i am not sure how dual-licensing with MIT and LGPL plays into this (see https://github.com/PenguinMod/PenguinMod-ExtensionsGallery/blob/main/static/extensions/Ashime/MoreFields.js#L5C4-L5C35 )

Basically due to how LGPL works the only thing that applies from the MIT license when using MIT AND LGPL (explicit for a reason) is the credit requirement and license notice

static/extensions/LICENSING.md
```diff
@@ -0,0 +1,9 @@
+# Licensing
+PenguinMod extension code in this folder is under one of two licenses:
+- [MIT](https://opensource.org/license/mit)
+- [LGPL-3.0](https://opensource.org/license/LGPL-3.0)
```
yuri-kiss: similar to what I said earlier

Steve0Greatness:
> only thing left is i am not sure how dual-licensing with MIT and LGPL plays into this (see https://github.com/PenguinMod/PenguinMod-ExtensionsGallery/blob/main/static/extensions/Ashime/MoreFields.js#L5C4-L5C35 )

The normal way that dual licensing works is the ability to change licenses as needed, for instance, dual licensing under MPL and AGPL would allow for the code to be AGPL in the case of server-sided-ness and MPL in the case of being used as a library. Usually the decision about which terms apply is left up to the user of the code as well, although there are some projects such as Redis which have used specific models where they decide which license you get (IE, you pay them for a less restrictive license).

The way that it works in More Fields is a very specific way that it works, and, to my knowledge, is entirely equivalent to just having it be under the LGPL, as the only real requirement of the MIT license is the requirement for the copyright header to be included, which the LGPL requires as well. It'd likely be a good idea for Yuri-Kiss to clarify this by editing the extension to simply be under LGPL.

I'd like to say that I have no issue with dual licensing, and I think that, like Yuri-Kiss requested above, it should be allowed, although maybe a slight clarification on what kinds of dual licensing are allowed. IMO, probably the user-choice modal is the kind that should be allowed, since that's the most inline with how others use dual licensing, and how people who are aware of dual licensing will likely expect it to work.

I'll take a quick look over the current file changes to see if there are any issues, but otherwise it should be fine.

static/extensions/LICENSING.md
```diff
@@ -0,0 +1,9 @@
...
+
+For an extension to use a specific license, it should specify the SPDX license code (seen above) in a comment at or near the top of the file.
```
Steve0Greatness: SPDX calls them "short identifiers."

README.md
```diff
@@ -1,5 +1,23 @@
...
+### Extension code licensing
+All PenguinMod extension code submitted to this repository is under
+one of two licenses:
+- [MIT](https://opensource.org/license/mit)
+- [LGPL-3.0](https://opensource.org/license/LGPL-3.0)
+
+For an extension to use a specific license, it should specify the SPDX license code (seen above) in a comment at or near the top of the file.
+
+If an extension does not specify a license, it is assummed to be released under the MIT license with
+the copyright holders being the people credited to said extension.
```
Steve0Greatness: Please be very sparing with repeating details across multiple files: it can be a pain to change later.

README.md
```diff
@@ -1,5 +1,23 @@
...
+
+Extension thumbnails and documentation are under the MIT license.
```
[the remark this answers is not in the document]
Steve0Greatness: I should probably note that by "specified somewhere" I mean somewhere else currently in the current version of the codebase, and not within this PR, I know this was earlier in this document of the PR.

Steve0Greatness (changes requested):

yuri-kiss:
> > only thing left is i am not sure how dual-licensing with MIT and LGPL plays into this (see https://github.com/PenguinMod/PenguinMod-ExtensionsGallery/blob/main/static/extensions/Ashime/MoreFields.js#L5C4-L5C35 )
> 
> The normal way that dual licensing works is the ability to change licenses as needed, for instance, dual licensing under MPL and AGPL would allow for the code to be AGPL in the case of server-sided-ness and MPL in the case of being used as a library. Usually the decision about which terms apply is left up to the user of the code as well, although there are some projects such as Redis which have used specific models where they decide which license you get (IE, you pay them for a less restrictive license).
> 
> The way that it works in More Fields is a very specific way that it works, and, to my knowledge, is entirely equivalent to just having it be under the LGPL, as the only real requirement of the MIT license is the requirement for the copyright header to be included, which the LGPL requires as well. It'd likely be a good idea for Yuri-Kiss to clarify this by editing the extension to simply be under LGPL.
> 
> I'd like to say that I have no issue with dual licensing, and I think that, like Yuri-Kiss requested above, it should be allowed, although maybe a slight clarification on what kinds of dual licensing are allowed. IMO, probably the user-choice modal is the kind that should be allowed, since that's the most inline with how others use dual licensing, and how people who are aware of dual licensing will likely expect it to work.
> 
> I'll take a quick look over the current file changes to see if there are any issues, but otherwise it should be fine.

I do not plan on editing it under just LGPL for numerous reasons (it is outdated code and theres no critical reason to change it)

"You must comply with the LGPL-3.0s terms regarding modifications and linking, while also respecting the MIT licenses attribution requirement." the attribution requirement in LGPL-3.0 works differently to the MIT one which is why it is there, the key difference between how I am using it is I am not using `MIT or` I am using `MIT AND/&`, its mixed licensing which is a type of dual licensing, jeremy said if this can be clarified in easy wording then its fine as is.

static/extensions/LICENSING.md
```diff
@@ -0,0 +1,9 @@
...
+
+If an extension does not specify a license, it is assummed to be released under the MIT license with
+the copyright holders being the people credited to said extension.
```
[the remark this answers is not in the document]
yuri-kiss: its a redundancy thing, and by default they already are MIT when not specified because they are under the site license when not given one

README.md
yuri-kiss: i feel like we can make a seperate pull request to split up the read me in a more orderly fashion

Steve0Greatness:
I would like to convey that the LGPL and the MIT License are, in regards to attribution, almost identical. The LGPL, as it acts as a modification on-top of the terms of the prior existing GPL, uses the same rules for attribution laid out in the MIT License. Just for clarification, here is the term in the MIT License that requires attribution:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

No matter what, you must...

* include a copyright notice;
* include a copy of the license with distribution.

The portions of the (L)GPL that regard attribution are far longer, so feel free to read the full, combined license to check me on this if you are so inclined <https://www.gnu.org/licenses/lgpl+gpl-3.0.txt> (to make it easier, the parts that cover this in particular are sections 4 to 6, although only sections 4 and 5 cover the specific use cases that are useful to this discussion). I'll boil it down here, though.

No matter what, you must...

* include the copyright notice;
* notify the user of the terms you received it under;
* convey the lack of warranty in the code of the project that is (L)GPL.
* include a copy of the license with distribution.

Notice how the terms of the LGPL cover everything that the MIT license does. The only thing that MIT appears to be doing here is requiring that a copy of the MIT license be distributed with the program, which is pointless because all terms of the MIT, and permissions given under it are fully covered by the LGPL.

yuri-kiss (approved):

PR #422, opened 2025-11-21
Lgpl licensing: my suggestions

yuri-kiss:
See: https://github.com/PenguinMod/PenguinMod-ExtensionsGallery/pull/419#issuecomment-3564066054


Steve0Greatness:
Note: they aren't all licensed under _both_ MIT and LGPL. You should change the wording to "and/or."

JeremyGamer13:
i dont quite like the placement of this but i think the wording is fine
the best version may be to merge this info with the last PR, and put it in a section before the full tutorial

PR #430, opened 2025-11-24
BIG update to the Chess extension

Gen1xLol:
yeah

### DIFF
--- a/static/extensions/Gen1x/chess-ext.js
+++ b/static/extensions/Gen1x/chess-ext.js
@@ -459,7 +459,8 @@
                     { opcode: 'hideBoard', blockType: 'command', text: 'hide chess board' },
                     { opcode: 'initializeDetector', blockType: 'Boolean', text: 'board initialized?' },
                     { blockType: 'label', text: 'Appearance & Settings' },
-                    { opcode: 'setBoardPosition', blockType: 'command', text: 'set board position to [POSITION]', arguments: { POSITION: { type: 'string', menu: 'positionMenu', defaultValue: 'center' } } },
+                    { opcode: 'setBoardPosition', blockType: 'command', text: 'set board position to [POSITION]', arguments: { POSITION: { type: 'string', menu: 'positionMenu', defaultValue: 'center' } }, hideFromPalette: true },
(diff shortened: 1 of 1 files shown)


src/lib/extensions.js
```diff
@@ -440,8 +440,8 @@ export default [
         isGitHub: false,
     },
     {
-        name: "Chess",
-        description: "A powerful extension about Chess. It allows to create a Chessboard, manage chess games, and use Stockfish 17 in your projects.",
+        name: "Chess (updated)",
```
RedMan13: remove the `(updated)` word, no point in including it

static/extensions/Gen1x/chess-ext.js
```diff
@@ -1,906 +1,1077 @@
...
+                    styleMenu: { items: ['emojis', 'icons'] },
+                    onOffMenu: { items: ['on', 'off'] },
+                    easingMenu: { items: ['linear', 'ease', 'ease-in', 'ease-out', 'ease-in-out'] },
+                    perspectiveMenu: { acceptReporters: true, items: ['white', 'black'] }
+                }
+            };
+        }
+        loadCpuEngine() {
+            if (this.cpuEngineLoaded) return;
+            fetch('https://rawcdn.githack.com/Gen1xLol/testassets/refs/heads/main/lozza.js')
```
RedMan13: it would probably be best if this returned the fetch promise so that the vm awaits this, that way you can verify if the load succeeded or not just by checking `is CPU engine loaded?` afterwards, rather then needing a whole separate block to be added.

JeremyGamer13: should use Scratch.fetch
also this can probably be fetched twice if you run the block quickly, not sure if that causes any bugs

JeremyGamer13 (changes requested):
checking blocks

Gen1xLol:
Fixed

RedMan13 (changes requested):
this is all, once these are all patched out theres no bugs or over sights i can see that make this an issue to merge

src/lib/extensions.js
RedMan13: ignore this, i didnt notice it got fixed

RedMan13:
i missed one!! tweening doesnt seem to actually do anything at all

Gen1xLol:
all done. we're ready for launch

PR #439, opened 2025-11-27
Added Both Revoke and Read function to the Blob Extension

faunks:
TheShovel, what's up pookie ;P

### DIFF
--- a/static/extensions/Faunks/Blobs.js
+++ b/static/extensions/Faunks/Blobs.js
@@ -21,7 +21,7 @@
           {
               opcode: "toBlob",
               blockType: Scratch.BlockType.REPORTER,
-              text: "Turn [DATA] to a [TYPE] blob",
+              text: "Turn [DATA] from [ENCODING] to a [TYPE] blob",
               arguments: {
                   DATA: {
                       type: Scratch.ArgumentType.STRING
@@ -31,20 +31,25 @@
                       defaultValue: "text/plain",
                       menu: "types"
                   },
+                  ENCODING: {
+                      type: Scratch.ArgumentType.STRING,
+                      defaultValue: "b64",
+                      menu: "encodings",
+                  }
               },
           },
           {
               opcode: "readBlob",
               blockType: Scratch.BlockType.REPORTER,
-              text: "Read blob [URL] as a [TYPE]",
+              text: "Read blob [URL] as [ENCODING]",
               arguments: {
                   URL: {
                       type: Scratch.ArgumentType.STRING
                   },
-                  TYPE: {
+                  ENCODING: {
                       type: Scratch.ArgumentType.STRING,
-                      defaultValue: "text/plain",
-                      menu: "types"
+                      defaultValue: "b64",
+                      menu: "encodings",
                   }
               },
           },
@@ -60,6 +65,15 @@
           },
         ]
         ,menus: {
+          encodings: {
+            acceptReporters: true,
+            items: [
+              { text: "Base64", value: "b64" },
+              { text: "Data URL", value: "dataurl" },
+              { text: "Plain Text", value: "plain" },
+              { text: "Byte Array", value: "array" }
+            ]
+          },
           types: {
             acceptReporters: true,
             items: [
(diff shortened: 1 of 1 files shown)


static/extensions/Faunks/Blobs.js
```diff
@@ -103,10 +128,40 @@
...
+        const response = await fetch(url);
+        const blob = await response.blob();
+        
+        if (!url.startsWith("blob:")) return "";
+
+        if (blob.size === 0) return ""; // js for u TheShovel
+
+        if (mime.startsWith("text/") || mime === "application/json") {
+            const text = await blob.text();
+            return mime === "application/json" ? text : text;
```
Steve0Greatness: Just return `text`, here.

static/extensions/Faunks/Blobs.js
```diff
@@ -33,6 +33,31 @@
                   },
               },
           },
+          {
+              opcode: "readBlob",
+              blockType: Scratch.BlockType.REPORTER,
+              text: "read blob [URL] as a [TYPE]",
```
Steve0Greatness: This casing on the reporter here is inconsistent with the other blocks.

static/extensions/Faunks/Blobs.js
```diff
@@ -103,10 +128,40 @@
...
+            return mime === "application/json" ? text : text;
+        }
+
+        const buffer = await blob.arrayBuffer();
+        const bytes = new Uint8Array(buffer);
+
+        let binary = "";
+        for (let i = 0; i < bytes.length; i++) {
+            binary += String.fromCharCode(bytes[i]);
+        }
```
Steve0Greatness: Could just be me, but I think there might be more efficient ways of converting an array buffer into a string. Maybe `TextDecoder`, or utilizing a `DataView` instead of `Uint8Array`. One thing that I know for a fact is that V8 (Chrome) handles `DataView` more efficiently than `Uint8Array`: https://v8.dev/blog/dataview. Not sure on the state of Firefox, here, though.

static/extensions/Faunks/Blobs.js
```diff
@@ -103,10 +128,43 @@
...
+    }
+    
+    revokeBlob(args, util){
+        URL.revokeObjectURL(args.URL);
+    }
+
+    readBlob(args, util) {
+    async function readBlobContent(url, mime) {
+        
+        if (!url.startsWith("blob:")) return ""; // I WOULD return "Url Error", but it would make it a bit harder to track if an output has been returned or not
```
Steve0Greatness: You could throw an error. So ``throw new URIError("must be a blob url");`` or something similar.

static/extensions/Faunks/Blobs.js
```diff
@@ -103,10 +128,43 @@
...
+        const blob = await response.blob();
+        
+
+        if (blob.size === 0) return ""; // js for u TheShovel
+
+        if (mime.startsWith("text/") || mime === "application/json") {
+            return await blob.text();
+        }
+
+        const buffer = await blob.arrayBuffer();
```
Steve0Greatness: After doing a bit of looking at the [MDN documentation for `Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob), there may be a way of doing this with a readable stream, although I'd need to check how efficient that is. Also quick note that the `text` method doesn't seem to require text to actually be text, so it can pretty much be anything and still be interpreted as text.

static/extensions/Faunks/Blobs.js
```diff
@@ -103,10 +128,43 @@
...
+        const view = new DataView(buffer);
+        
+        let binary = ''; // = decoder.decode(bytes); Decoder doesn't work with btoa, and replacing it reduces efficiency.
+        
+        for (let i = 0; i < view.byteLength; i++) {
+          binary += String.fromCharCode(view.getUint8(i));
+        }
+        return btoa(binary);
```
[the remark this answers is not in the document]
Steve0Greatness: Also, I think that outputting it as b64 could be considered unexpected behavior, mostly because input isn't given in base 64.

static/extensions/Faunks/Blobs.js
Steve0Greatness: I could also be wrong about this.

static/extensions/Faunks/Blobs.js
yuri-kiss: > Could just be me, but I think there might be more efficient ways of converting an array buffer into a string. Maybe `TextDecoder`, or utilizing a `DataView` instead of `Uint8Array`. One thing that I know for a fact is that V8 (Chrome) handles `DataView` more efficiently than `Uint8Array`: https://v8.dev/blog/dataview. Not sure on the state of Firefox, here, though.

That blog article says otherwise? "We found that our new DataView implementation provides **almost the same performance** as TypedArrays when accessing data aligned in the native endianness (little-endian on Intel processors)"

static/extensions/Faunks/Blobs.js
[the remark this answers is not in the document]
yuri-kiss: unless im misreading your phrasing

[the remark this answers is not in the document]
RedMan13: wgat

Steve0Greatness:
I would like to note that since Array Buffers has been merged, this extension would actually work fairly well with array buffers. It has the ability to encode and decode every single encoding option provided by this extension, meaning it would be useful for people who want to read from and write to blobs.

image

Steve0Greatness:
> (I didn't)

Yeah, I'm aware. The 4 formats are some of the only ones that come to mind quickly for serializing binary data, so it's not unforeseen that they'd turn out to be the formats chosen in both cases.

PR #446, opened 2025-12-11
PenguinMic

dot-jms:
PenguinMic is an extension that lets you record and play back audio from your microphone in PenguinMod.


Steve0Greatness:
That's not something that can be particularly trusted. Small issues or feature requests come in all the time for things that you might not have even thought of.

PR #451, opened 2025-12-27
add server storage extension

Ikelene:
idk why it still wants to merge update url and comments (it was already merged) 
and i did sync

anyways my new super cool extension is here
server is open source at https://github.com/Ikelene/scratchStorageExtension
thanks

### DIFF
--- a/static/extensions/Ikelene/serverStorageExtension.js
+++ b/static/extensions/Ikelene/serverStorageExtension.js
@@ -42,9 +42,9 @@
             return {
                 id: 'ikeleneServerStorage',
                 name: 'Server Storage',
-                color1: '#ff9bfd',
-                color2: '#ff9bfd',
-                color3: '#ff9bfd',
+                color1: '#ff6bfd',
+                color2: '#ff6bfd',
+                color3: '#ff6bfd',
                 docsURI: this.serverUrl + 'docs.html',
                 blocks: [
                     {


static/extensions/Ikelene/serverStorageExtension.js
```diff
@@ -0,0 +1,445 @@
...
+                return 'need api key';
+            }
+            try {
+                const res = await fetch(this.serverUrl + 'listKeys.php', {
+                    method: 'POST',
+                    headers: {
+                        'Content-Type': 'application/json',
+                        'Accept': 'application/json'
+                    },
+                    body: JSON.stringify({ apiKey: this.apiKey })
```
Steve0Greatness: This would require changes to the server, but in general API key sending should be done through headers, such as [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Authorization) using the basic method. If this is done, then the request could instead be sent as using the GET method, which better explains what the actual request is doing.

static/extensions/Ikelene/serverStorageExtension.js
```diff
@@ -0,0 +1,445 @@
...
+        }
+
+        setEditMode(args) {
+            const mode = args.MODE === 'local' ? 'local' : 'live';
+            this.editMode = mode;
+        }
+
+        async getAllKeys() {
+            if (!this.hasValidKey()) {
+                return 'need api key';
```
Steve0Greatness: Every instance of this should be throwing rather than warning and returning, or returning a string. This could very easily lead to unnoticed bugs.

static/extensions/Ikelene/serverStorageExtension.js
```diff
@@ -0,0 +1,445 @@
...
+                });
+
+                const result = await response.json();
+                if (!result.success) {
+                    console.warn('You need to set an API key first. You can generate a key if you press the Open Documentation button');
+                }
+            } catch (error) {
+                console.warn('You need to set an API key first. You can generate a key if you press the Open Documentation button');
+            }
+        }
```
Steve0Greatness: An error could occur for more reasons than just the API key not being set... not to mention that I'm pretty sure there's not really a way for a user to have gotten to this point without having set one.

static/extensions/Ikelene/serverStorageExtension.js
Steve0Greatness: API key should also be sent in headers of other data posting queries.

static/extensions/Ikelene/serverStorageExtension.js
```diff
@@ -0,0 +1,445 @@
...
+                    },
+                    body: JSON.stringify({
+                        apiKey: this.apiKey,
+                        key: key
+                    })
+                });
+
+                if (response.status === 401) {
+                    return 'You need to set an API key first. You can generate a key if you press the Open Documentation button';
+                }
```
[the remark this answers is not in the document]
Steve0Greatness: Also, throw

static/extensions/Ikelene/serverStorageExtension.js
```diff
@@ -0,0 +1,445 @@
...
+                    body: JSON.stringify({
+                        apiKey: this.apiKey,
+                        key: key
+                    })
+                });
+
+                const result = await response.json();
+                if (!result.success) {
+                    console.warn('Delete failed or key not found');
+                }
```
Steve0Greatness: Maybe send an error message saying what went wrong? In addition to appropriate status codes, of course. 500 should be sent in the case of a failure to delete a key. Also, here would be a good place to throw.

static/extensions/Ikelene/serverStorageExtension.js
```diff
@@ -0,0 +1,445 @@
...
+                return '[]';
+            }
+        }
+
+        async downloadCache() {
+            if (!this.hasValidKey()) {
+                console.warn('need api key to download cache');
+                return;
+            }
+            try {
```
Steve0Greatness: In general, if the function need to be able to perform networking and is unable to, it should be able to error out. Here it makes sense to dump the responsibility to handle it onto the user of the extension.

static/extensions/Ikelene/serverStorageExtension.js
```diff
@@ -0,0 +1,445 @@
...
+        }
+
+        getInfo() {
+            return {
+                id: 'ikeleneServerStorage',
+                name: 'Server Storage',
+                color1: '#ff9bfd',
+                color2: '#ff9bfd',
+                color3: '#ff9bfd',
+                docsURI: this.serverUrl + 'apiKey.html',
```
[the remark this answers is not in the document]
Steve0Greatness: On such an extension docs, I'd recommend including a short section about hosting your own server. In addition to an explanation of the purpose of API keys, and the differences between the editing modes.

static/extensions/Ikelene/serverStorageExtension.js
```diff
@@ -0,0 +1,445 @@
...
+                    })
+                });
+
+                if (response.status === 401) {
+                    return 'You need to set an API key first. You can generate a key if you press the Open Documentation button';
+                }
+
+                const result = await response.json();
+                if (result.success && result.data) {
+                    return result.data.value || '';
```
Steve0Greatness: This API is weirdly complicated. You may want to trim back what it sends in the main body a little bit.

static/extensions/Ikelene/serverStorageExtension.js
```diff
@@ -0,0 +1,445 @@
...
+
+            if (this.editMode === 'local') {
+                delete this.localCache[key];
+                this.dirtyKeys.add(key); // optional: also push delete server-side later via dedicated delete api
+                return;
+            }
+
+            try {
+                const response = await fetch(this.serverUrl + 'delete.php', {
+                    method: 'POST',
```
Steve0Greatness: There's actually a specific [`DELETE`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/DELETE) method made for the exact purpose. It's even allowed to have a request body.

cicerorph:
Btw, just here to say that If you didn't add protection on the php script for setting data, ADD IT, people can just go there and prob find a way to create a php file and run it, and as you know, it gives access to the server, just sayin btw, gn

Ikelene:
Hey can we do something about people placing their extensions at the top of the list? My extension pretty much just got replaced by another one, and that one was out at the very top of the list, making mine obsolete because theirs gets all the exposure and mine will never get seen because who wants to scroll down when an extension that fits their needs is right there. Maybe have it so there's tags for extensions like #auth or #storage etc to sort them and give all extensions equal exposure. Also it's kind of unfair because my extension PR has been open for 3 weeks now and theirs got approved within a day of opening :/

JeremyGamer13:
> Hey can we do something about people placing their extensions at the top of the list?

we reorder the list every now and then but PRs to reorder should be fine too

> Maybe have it so there's tags for extensions like #auth or #storage etc to sort them and give all extensions equal exposure.

will be solved by #482

static/extensions/Ikelene/serverStorageExtension.js
JeremyGamer13: given that even PM doesn't do this, i think its fine for a merge

JeremyGamer13 (changes requested):
this is basically the only issue i have with this

PR #453, opened 2025-12-29
Add PangAI extension

Seigh-sword:
PangAI-Icon 

add Pang AI extension and documentation, please


jwklong:
https://github.com/PenguinMod/PenguinMod-ExtensionsGallery?tab=readme-ov-file#important-guidelines
> Your extension CANNOT rely on the use of any generative AI. This is mainly for extensions that respond to messages from AI like ChatGPT. Any extensions added before this rule are an exception.

PR #463, opened 2026-01-02
New Plugin (Full Pull Request)

REALMarrtyFDGU:


static/extensions/AeroHutch/date-to-hex.js
```diff
@@ -0,0 +1,73 @@
...
+          }
+        ]
+      };
+    }
+
+    dateToHex(args) {
+      const date = new Date(args.DATE);
+      const timestamp = date.getTime();
+      
+      if (isNaN(timestamp)) return "Invalid Date";
```
Steve0Greatness: If there's an issue that cannot be continued from, you should throw, not return. Throw will render an error message on the GUI and highlight the script that had an issue, returning doesn't.

static/extensions/AeroHutch/date-to-hex.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    hexToDate(args) {
+      try {
+        const hex = args.HEX.replace(/\s/g, '');
+        const bytes = new Uint8Array(hex.match(/.{1,2}/g).map(byte => parseInt(byte, 16)));
+        const view = new DataView(bytes.buffer);
+        
+        // Read as Big-Endian 64-bit integer
+        const timestamp = Number(view.getBigUint64(0, false));
```
Steve0Greatness: Yet again, I think this entire area could fairly well replaced with a singular call to `parseInt`, with the second parameter set to 16, to make sure it knows to parse as a 16 bit integer.

static/extensions/AeroHutch/date-to-hex.js
Steve0Greatness: Please be consistent about your usage of single quotes (`'`) or double quotes (`"`) for strings in this file. There really isn't a difference between the two (not in JavaScript, at least), you can easily interchange them. Given you use single quotes most often, I'd replace the double quotes in this file with single quotes.

static/extensions/AeroHutch/date-to-hex.js
```diff
@@ -0,0 +1,73 @@
...
+      
+      if (isNaN(timestamp)) return "Invalid Date";
+
+      // Create an 8-byte buffer
+      const buffer = new ArrayBuffer(8);
+      const view = new DataView(buffer);
+
+      // Write as Big-Endian 64-bit integer
+      // Note: BigInt is used because JS numbers lose precision above 53 bits
+      view.setBigUint64(0, BigInt(timestamp), false);
```
Steve0Greatness: Yes. Big endian is the default mode of this method, anyway. Specifying it to be written in big endian in a bit pointless: [`DataView.prototype.setBigUint64()` `littleEndian`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/setBigUint64#littleendian).

Steve0Greatness (changes requested):

PR #470, opened 2026-01-10
Add Cursor Controls by codemaster

moosanaeempc-lgtm:
With this extension you can lock and hide your pointer while tracking your mouse movement best to use in 3d games.


moosanaeempc-lgtm:
It don't works when mouse with lock.

Steve0Greatness:
> It don't works when mouse with lock.

What are you even talking about? The sensing mouse x and mouse y blocks _do_ report difference in mouse position when the mouse is locked with Pointerlock. This is documented behavior. If it isn't working then that should be a bug report with TurboWarp not an entirely new extension.

PR #471, opened 2026-01-15
Add AES Encryption Extension

2015kaito8169:
Extension Name: AES
Feature: This extension allows users to encrypt and decrypt strings using the AES (Advanced Encryption Standard).
It supports both Japanese and English characters (UTF-8).
Easy to use for any project that needs secure data.
Perfect for securing cloud variables, save data, or secret messages.
I've added the JS file and registered it in the extension list. Please review and merge it. Thank you!
 
bykaito8169

### DIFF
--- a/src/lib/extensions.js
+++ b/src/lib/extensions.js
@@ -548,8 +548,7 @@ export default [
         isGitHub: true,
         creatorAlias: "gaimerI17",
     },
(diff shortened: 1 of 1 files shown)


LordCat0:
The text on the blocks and the extension name should be in english

PR #485, opened 2026-01-26
Storage Plus

Gen1xLol:
yea
codename was "Betterer Storage" but on second thought that name sucks


Gen1xLol:
i removed the toggle lock block, i forgot i added it honestly (it was just there for testing)

PR #492, opened 2026-02-03
Add Move Controls extension for Scratch

NiCaso14:


Ianyourgod:
We do not allow extensions made with Extension Forge or any other similar tools to be added to the gallery.

PR #506, opened 2026-02-07
Regular Expressions: Clone Block

DogeisCut:
vivaldi_5TNpbjidvt
vivaldi_DAyT0Bu1VP
thats it

### DIFF
--- a/src/lib/Extension/Component.svelte
+++ b/src/lib/Extension/Component.svelte
@@ -348,7 +348,7 @@
         display: flex;
         flex-direction: column;
         justify-content: space-between;
-        
+
         border-radius: 6px;
     }
     :global(body.dark-mode) .block {
--- a/src/lib/extension-tags.js
+++ b/src/lib/extension-tags.js
@@ -29,15 +29,20 @@ export const Tags = [
         alias: "Extension Collections",
         group: "extensiontypes",
     },
-
-    // any other misc ones that just need aliases
     {
         name: "customtype",
         alias: "New Block Type",
+        group: "extensiontypes",
     },
+
+    // any other misc ones that just need aliases
     {
         name: "genai",
         alias: "Generative AI",
+    },
+    {
+        name: "api",
+        alias: "API / Other websites",
     },
     {
         name: "ai",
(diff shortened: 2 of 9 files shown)


DogeisCut:
> image
> this is the array version. It's caused by the array extension loading *after*. The iterators ext doesn't have this issue so do what it does I guess.
> 
> Also, loading the iterator ext after loading this extension won't make the iterator version of the block appear. Not sure if this is possible to be fixed.

thats because there isnt an iterators version of the block. i made this extension before iterators was finished and never got around to it

as for the loading arrays after... you arent supposed to even be able to do that because this extension is supposed to autoload arrays...

DogeisCut:
> > image
> > this is the array version. It's caused by the array extension loading *after*. The iterators ext doesn't have this issue so do what it does I guess.
> > 
> > Also, loading the iterator ext after loading this extension won't make the iterator version of the block appear. Not sure if this is possible to be fixed.
> 
> thats because there isnt an iterators version of the block. i made this extension before iterators was finished and never got around to it
> 
> as for the loading arrays after... you arent supposed to even be able to do that because this extension is supposed to autoload arrays...

oh nvm guess i was working on one and forgot???

Ianyourgod:
what i said is that arrays gets AUTOLOADED after. i am not loading it myself. the bug do not occur if i loaded it myself prior to loading this extension.

DogeisCut:
this bug doesnt happen on the current version of regex at least for me so idk what i did

static/extensions/DogeisCut/dogeiscutRegularExpressions.js
```diff
@@ -424,8 +432,21 @@
                                 defaultValue: "foo"
                             },
                         },
-                        hideFromPalette: !vm.runtime.ext_jwArray,
-                        ...(vm.runtime.ext_jwArray ? vm.jwArray.Block : {}),
+                        hideFromPalette: !vm.runtime.ext_jwArray && vm.runtime.ext_divIterator,
+                        ...(vm.runtime.ext_jwArray && !vm.runtime.ext_divIterator ? vm.jwArray.Block : {}),
+                    },
+                    {
+                        opcode: 'matchAllIterator',
```
SharkPool-SP: no block definition?

PR #507, opened 2026-02-11
Objects: Complete Rewrite

DogeisCut:
# Changes
- is faster
- everything is compiled
- uses maps instead of objects
- like 2 new blocks
- hides deprecated block
- way less buggy
- table display works with arrays properly
- cleaner code that doesnt make me cry when i go to edit it
- lots of other small cool stuff
 
# Images
image
image
image

### DIFF
--- a/static/extensions/DogeisCut/dogeiscutObject.js
+++ b/static/extensions/DogeisCut/dogeiscutObject.js
@@ -537,7 +537,8 @@
             vm.dogeiscutObject = dogeiscutObject
             vm.runtime.registerSerializer(
                 "dogeiscutObject",
-                mapType => Object.fromEntries(Array.from(mapType.map).map(([key, value]) => {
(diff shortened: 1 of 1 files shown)


DogeisCut:
how do i add reviewers

DogeisCut:
Opening the PR from a draft because nobody has found any bugs with it so far and it's fairly ready to be merged. (Would be nice If I could add some more block switches but the block switching API hates object/array inputs and errors out... I cant get a block with an array input to switch to/from another block with no array input)

DogeisCut:
> it seems like objects set into variables from the old version are broken in the new version
> 
> old: image image
> 
> new: image image
> 
> project file in discord: https://discord.com/channels/1033551490331197462/1243430348499451976/1471269738607808756
> 
> everything else seems fine

fixed

PR #520, opened 2026-02-13
Housekeeping on ProjectInterfaces.js

LordCat0:
Multiple people have pointed out how bad my coding was back then when I first made this.
I can obviously do better now.
This will format the extension better, and fix a few bugs and performance issues.


static/extensions/LordCat0/ProjectInterfaces.js
```diff
@@ -3,602 +3,1286 @@
...
+                        opcode: "Direction",
+                        text: "set direction of id [id] to [dir]",
+                        blockType: Scratch.BlockType.COMMAND,
+                        arguments: {
+                            id: {
+                                type: Scratch.ArgumentType.STRING,
+                                defaultValue: "My element",
+                            },
+                            dir: {
+                                type: Scratch.ArgumentType.NUMBER,
```
JeremyGamer13: maybe this should use the ANGLE picker?

JeremyGamer13 (changes requested):
cant review the whole thing right now but i have 2 notes

PR #559, opened 2026-04-10
Temp depr scopevar

yuri-kiss:
fixes: #557
also hides the extension so it doesn't cause more issues

this extension has a lot of debt as it is from an older time where there was no good way to patch the compiler or do what it does, and as such it needs to be rewritten, which I will do soon


yuri-kiss:
@Steve0Greatness

yuri-kiss:
better diff for that minified code: (I manually wrote in this fix)
image

Steve0Greatness (approved):
This works for the time being.

PR #562, opened 2026-04-15
Date Format V2: Fix Seriouslization

ddededodediamante:
By Jove, it seems that serialization was broken all this time!


LordCat0:
ddededodediamante

PR #571, opened 2026-05-01
Make Array Buffers Compiled

AndrewScripter587:
This PR makes the array buffers extension compiled, and adds a couple blocks.


AndrewScripter587:
Why are commits from months ago here
eh whatever

PR #573, opened 2026-05-10
pretty please add my nodes extension

ZiploxWasTaken:
its really cool trust!!!

fix if its broken tho...


static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,397 @@
...
+        opcode: "allLinks",
+        blockType: Scratch.BlockType.REPORTER,
+        hideFromPalette: false,
+        color1: LINK1,
+        color2: LINK2,
+        color3: LINK3,
+        text: "all links",
+        arguments: {},
+        disableMonitor: true
+    });
```
Steve0Greatness: Don't do this. Just put all of these into one array one after the other and directly put it into the return of the `getInfo` method.

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,397 @@
...
+    class Extension {
```
Steve0Greatness: You should probably rename this class to something more descriptive than "Extension."

src/lib/Documentation/Node-System.md
Steve0Greatness, line 1: When doing documentation, it should actually explain what the extension does.

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,397 @@
...
+                    }
+                }
+            };
+        }
+createNode(args) {
+    nodes.push({
+        id: Number(args.ID),
+        x: Number(args.X),
+        y: Number(args.Y)
+    });
```
Steve0Greatness: Given the way you're using nodes, this would be far more efficient as a map rather than a list.

With a Map you can get near O(1) time for querying by id, whereas here querying by id is O(n) time.

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,397 @@
...
+}
+        deleteNode(args) {
+            const id = Number(args.ID);
+            const index = nodes.findIndex(
+                n => n.id === id
+            );
+            if (index !== -1) {
+                nodes.splice(index, 1);
+            }
```
Steve0Greatness: This could be reduced to one line if nodes were a Map.

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,397 @@
...
+            const b = Number(args.B);
+            for (let i = links.length - 1; i >= 0; i--) {
+                const l = links[i];
+                if (
+                    (l.from === a && l.to === b) ||
+                    (l.from === b && l.to === a)
+                ) {
+                    links.splice(i, 1);
+                }
+            }
```
Steve0Greatness: Arrays have a filter method.

static/extensions/ZiploxWasTaken/nodes.js
```diff
@@ -0,0 +1,397 @@
...
+                n => n.x === x && n.y === y
+            );
+            return found ? found.id : "";
+        }
+        linkNodes(args) {
+            const a = getNode(args.A);
+            const b = getNode(args.B);
+            if (!a || !b) return;
+            if (a.id === b.id) return;
+            if (!areAdjacent(a, b)) return;
```
Steve0Greatness, line 665: Does this limit linking nodes to nodes that are only 1 apart?

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,397 @@
...
+    "use strict";
+    if (!Scratch.extensions.unsandboxed) {
+        alert("This extension must be unsandboxed.");
+        return;
+    }
+    const nodes = [];
+    const links = [];
+    const blocks = [];
+    const menus = {};
+    let nextId = 1;
```
Steve0Greatness: Unused variables.

Steve0Greatness (changes requested):

static/extensions/EverythingAcc0unt/nodes.js
ZiploxWasTaken: how was I supposed to know that

static/extensions/EverythingAcc0unt/nodes.js
ZiploxWasTaken: you can rename this?

src/lib/Documentation/Node-System.md
ZiploxWasTaken: leave it unused

static/extensions/EverythingAcc0unt/nodes.js
ZiploxWasTaken: I dont know what maps are

static/extensions/EverythingAcc0unt/nodes.js
ZiploxWasTaken: like I said I don't know what maps are

static/extensions/EverythingAcc0unt/nodes.js
ZiploxWasTaken: this is literally my first extension

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,397 @@
...
+                n => n.x === x && n.y === y
+            );
+            return found ? found.id : "";
+        }
+        linkNodes(args) {
+            const a = getNode(args.A);
+            const b = getNode(args.B);
+            if (!a || !b) return;
+            if (a.id === b.id) return;
+            if (!areAdjacent(a, b)) return;
```
ZiploxWasTaken: its a feature!

static/extensions/EverythingAcc0unt/nodes.js
Steve0Greatness: You can initialize an array with entries in JavaScript.

Steve0Greatness: It's pretty similar to how you can initialize objects with keys and values.

static/extensions/EverythingAcc0unt/nodes.js
Ianyourgod: then learn. the point of a review is not to just tell you that you did something wrong and that be the end of it, the point of it is to tell you that the code can be improved or that there's an issue you need to fix before the code can be merged.

src/lib/Documentation/Node-System.md
RedMan13: > leave it unused

since the doc isnt used it really shouldnt be included

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,400 @@
...
+    const links = [];
+const NODE1 = "#56ad57";
+const NODE2 = "#419b42";
+const NODE3 = "#348235";
+
+const LINK1 = "#2f8242";
+const LINK2 = "#256e35";
+const LINK3 = "#1c5729";
+    function getNode(id) {
+    return nodes.get(Number(id));
```
RedMan13: Use `Scratch.Cast.toNumber`, the js number cast doesnt support many scratch notations like NaN being equal to zero

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,400 @@
...
+                color1: "#f06681",
+                color2: "#d6516c",
+                blocks: [
+
+            {
+                opcode: "createNode",
+                blockType: Scratch.BlockType.COMMAND,
+                hideFromPalette: false,
```
RedMan13: this is implied false, you dont need to include it

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,400 @@
...
+                    Y: {
+                        type: Scratch.ArgumentType.NUMBER,
+                        defaultValue: 0
+                    },
+                    ID: {
+                        type: Scratch.ArgumentType.NUMBER,
+                        defaultValue: 1
+                    }
+                },
+                disableMonitor: true
```
RedMan13: because this block contains inputs, this is implied true, so you dont need this either
though it really isnt an issue to keep it anyways

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,400 @@
...
+    nodes.delete(id);
+
+    for (let i = links.length - 1; i >= 0; i--) {
+        if (
+            links[i].from === id ||
+            links[i].to === id
+        ) {
+            links.splice(i, 1);
+        }
+    }
```
RedMan13: this can be reduced to just 
```js
const linkIdx = links.findIndex(link => link.from === id || link.to === id);
links.splice(linkIdx, 1);
```
also, for speed, it should be better to have each node store its target, and possibly even child, instead of needing to search in a list of links external to the list of nodes.
also would reduce the chance for errors to occur from any kind of state desync, where the list of nodes doesnt have a node referenced in the links.
also also, if you really want to put yourself to the hardest method, you can just store all nodes in an ordered list, where the item immediately before and immediately after are the to and from nodes.

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,400 @@
...
+                case "left":
+                    x -= 1;
+                    break;
+                case "up":
+                    y += 1;
+                    break;
+                case "down":
+                    y -= 1;
+                    break;
+            }
```
RedMan13: this fails if the nodes arent exclusively one-away from each other and perfectly on grid with eachother, maybe instead limit it to only `forward` and `backward`? or compute the angle to each node and find the one closest to an angle the user inputs.

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,398 @@
...
+            for (const found of nodes.values()) {
+    if (found.x === x && found.y === y) {
+        return found.id;
+    }
+}
+
+return "";
```
Steve0Greatness: You can use the iterator the values method returns has a find method. You could refactor this to use that:

Obviously you don't need to use that verbatim, this is just a suggestion of how you could refactor it.

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,398 @@
...
+            if (linkExists(a.id, b.id)) return;
+            links.push({
+                from: a.id,
+                to: b.id
+            });
+        }
+        unlinkNodes(args) {
+            const a = Number(args.A);
+            const b = Number(args.B);
+            //stevie shut the hell up dude!!!!!!1!//
```
Steve0Greatness: This is such an unneeded comment.

static/extensions/EverythingAcc0unt/nodes.js
RedMan13: > this fails if the nodes arent exclusively one-away from each other and perfectly on grid with eachother, maybe instead limit it to only `forward` and `backward`? or compute the angle to each node and find the one closest to an angle the user inputs.

or really actually, maybe just move this to links? that way it can be a link follower instead, which would be far more handy

static/extensions/EverythingAcc0unt/nodes.js
```diff
@@ -0,0 +1,398 @@
...
+
+    for (let i = links.length - 1; i >= 0; i--) {
+        if (
+            links[i].from === id ||
+            links[i].to === id
+        ) {
+            links.splice(i, 1);
+        }
+    }
+}
```
Steve0Greatness: Another spot you could use the filter method.

Steve0Greatness (changes requested):
The indenting is a bit of a mess. I'd probably run this through some kind of formatter as it's kinda hard to read.

RedMan13 (changes requested):
Pretty pretty pretty please make it so links and nodes can be tracked directly, right now theres no way from this extension to see which nodes are linked where, nor easily follow those links in any meaningful way

static/extensions/EverythingAcc0unt/nodes.js
Steve0Greatness: No wait that codeblock is awful

Steve0Greatness:
```js
return nodes.values().find(n => !(n.x === x && n.y === y))?.id ?? "";
```

RedMan13: > ```js
> return nodes.values().find(n => !(n.x === x && n.y === y))?.id ?? "";
> ```

isnt this wrong, since its just `x1 !== x2 || y1 !== y2`, but the original code is `x1 === x2 && y1 === y2`

static/extensions/EverythingAcc0unt/nodes.js
Steve0Greatness: That's within the deleteNode method. I think the intention is that it deletes every single link that includes each the nodes, rather than to only delete one link, which this would do.

RedMan13: > That's within the deleteNode method. I think the intention is that it deletes every single link that includes each the nodes, rather than to only delete one link, which this would do.

ngl up until like half way through i had the preconception that you could only have one link per node

RedMan13: but yes, marking this as resolved because its evil and invalid

static/extensions/EverythingAcc0unt/nodes.js
Steve0Greatness:
```js
return nodes.values().find(n => n.x === x && n.y === y)?.id ?? "";
```

Steve0Greatness: Real

static/extensions/EverythingAcc0unt/nodes.js
RedMan13: thisnisnt resolved, please actually engage before resolving

ZiploxWasTaken: it being only 1 grid space away is a feature but if you want I can find a way to make it be more than 1 grid space in a setting

src/lib/extensions.js
```diff
@@ -20,6 +20,16 @@ export default [
...
     },
+    {
+        name: "Nodes",
+        description: "A node system that can be utilized for many purposes, such as world maps and more!",
+        tags: ["new"],
+        code: "ZiploxWasTaken/nodes.js",
+        documentation: "nodesystems",
+        banner: "ZiploxWasTaken/Nodes.svg",
+        isGitHub: true,
+        creator: ["ZiploxWasTaken", "outsideflight"],
```
Steve0Greatness: This property currently doesn't have the ability to take an array.

static/extensions/ZiploxWasTaken/nodes.js
Steve0Greatness, line 1: Second time I'll say something about this. This file's formatting is very messy, particularly in the indenting department.

static/extensions/ZiploxWasTaken/nodes.js
```diff
@@ -0,0 +1,699 @@
...
+            targetY += 1;
+            break;
+        case "down":
+            targetY -= 1;
+            break;
+    }
+
+    const target = [...nodes.values()].find(
+        n => n.x === targetX && n.y === targetY
+    );
```
Steve0Greatness: I'm pretty sure you can use find on iterators, and don't need to expand them into arrays.

src/lib/Documentation/Node-System.md
```diff
@@ -0,0 +1,50 @@
+# Node System
+
+Hi! So... Just wanna let you know... this is actually the *second* version of the document. The first one got deleted because I'm pretty sure it was meant to be unused... This is OutsideFlight writing this, by the way. 
+
+Now, you clicked on this markdown so you could learn about our extension about nodes and links, right? Well, without furthur ado, here's the facts!
+
+<h2>Introduction</h2>
+
+The extension itself is pretty self explanatory, but if you need to know how every block works... this is the place to go.
```
Steve0Greatness: Explain it anyway.

static/extensions/ZiploxWasTaken/nodes.js
```diff
@@ -0,0 +1,699 @@
...
+    let id = 1;
+
+    for (let y = maxY; y >= minY; y--) {
+        for (let x = minX; x <= maxX; x++) {
+
+            nodes.set(id, {
+                id,
+                x,
+                y
+            });
```
Steve0Greatness: If there are already nodes using these ids, they're going to get overwritten.

This could allow for links to be created to nodes that are more than 1 away by creating 2 nodes (1 and say 100) and placing them directly adjacent to each other, then creating a grid that is 10 by 10 would have node 100 and node 1 be placed very far away from one another diagonally.

Steve0Greatness (changes requested):

src/lib/extensions.js
```diff
@@ -1,4 +1,4 @@
-/*
+,/*
```
Steve0Greatness: I think you accidentally added a comma here.

Steve0Greatness (changes requested):
Maybe the documentation needs a bit more work, but this is definitely better.

src/lib/extensions.js
ZiploxWasTaken: shit

Steve0Greatness (changes requested):
Seems like all the issues I had with this have been fixed.

Steve0Greatness (approved):

static/extensions/EverythingAcc0unt/nodes.js
RedMan13: still not resolved

AndrewScripter587:
Pretty cool extension, would be better if it used objects and arrays instead of json though

AndrewScripter587:
and also a way to load nodes/connections from an array too would be nice

ZiploxWasTaken:
I think it already uses arrays and objects to begin with... and you can use the backpack feature in the penguinmod editor or make a custom block to save your stuff. I need to look into the second one though.

AndrewScripter587:
I can tell you the 2nd one is false, the backpack only saves the blocks/sprite/costume/sound and the extension(s) themselves with no ext data saved

ZiploxWasTaken:
then you have to import the extension back

jwklong:
> I think it already uses arrays and objects to begin with... and you can use the backpack feature in the penguinmod editor or make a custom block to save your stuff. I need to look into the second one though.

arrays and objects as in the extensions

AndrewScripter587:
> > I think it already uses arrays and objects to begin with... and you can use the backpack feature in the penguinmod editor or make a custom block to save your stuff. I need to look into the second one though.
> 
> arrays and objects as in the extensions

Yeah that's what I meant, not json stuff

PR #581, opened 2026-05-16
Fix some stuff

AndrewScripter587:
mainly refresh the monitor (?) when the buffer changes and make the read block not stupid

### DIFF
--- a/static/extensions/AndrewGaming587/agBuffer.js
+++ b/static/extensions/AndrewGaming587/agBuffer.js
@@ -223,12 +223,14 @@
             if (typeof bi === "bigint") return bi;
             return BigInt(bi);
         }
+        _monitorUpToDate = false
         set(type = "Uint8",index,value,endian = false) {
             let buffer = this;
             // console.log(type)
             // console.log(index)
             // console.log(endian)
             // console.log(value)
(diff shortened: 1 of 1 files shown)


static/extensions/AndrewGaming587/agBuffer.js
```diff
@@ -1488,48 +1508,48 @@
                         // console.log(datatype)
                         let isConst = false
                         // console.log(datatype)
-                        compiler.source += `let ${bufName} = ${buffer};`
+                        compiler.source += `let ${bufName} = vm.agBuffer.Type.cast(${buffer};`
```
RedMan13: oopsies, missed the `)`

RedMan13 (changes requested):
otherwise this is entirely correct as i can tell

static/extensions/AndrewGaming587/agBuffer.js
AndrewScripter587: oops

AndrewScripter587:
> otherwise this is entirely correct as i can tell

fixed

AndrewScripter587 (commented):
review again please

PR #585, opened 2026-05-22
Add clicker extension with basic functionality

mln1238:
extension


SharkPool-SP:
Made with turbobuilder or whatever it's called. Not allowed. Same with the previous PR which is a duplicate.

PR #588, opened 2026-06-01
MIDI EXTENSION YAY

FloppyDisk-OSC:
You should like
update midi.js

adds joystick stuff


LincolnXGames:
so midiful

asterpawz:
He made a midi so good even his extension praised him

PR #590, opened 2026-06-08
Remove "custom CSS" block from Project Interfaces docs

LordCat0:
The custom CSS block was hidden in the extension recently and it doesn't make sense for it to be in the documentation anymore.


LordCat0:
I dont know why it added the first two commits

PR #596, opened 2026-06-16
Achievements

ZiploxWasTaken:
uyicjv bkjlgyifucvjhb.khuogyfvbkjlki/guybnj


cicerorph:
I know that like, this is kinda of a good idea but this is really easy to bypass getting the achievement soo

Also why do we have an extension for everything, let people make stuff

Steve0Greatness:
TBF, a lot of achievement systems are easy to bypass, it's just that it's not rewarding to, which is the primary reason that most people don't, despite it being fairly easy (except for certain cases where achievements may include actual rewards).

This is missing a bit of common functionality. Many achievement systems include achievements that require you to do *x* *y* times or find *z*, several of which may be scattered throughout the game. Just for one example, a game that I've been playing recently has a lot of collectable sandwiches found throughout its gameworld, each of which found increments a counter on an achievement.

static/extensions/ZiploxWasTaken/achievements.js
```diff
@@ -0,0 +1,445 @@
...
+    return { text: Scratch.translate(text), value };
+  }
+
+  loadRegistryFromStorage();
+  injectStyles();
+
+  class AchievementEngine {
+    getInfo() {
+      return {
+        id: "achievementengine",
```
JeremyGamer13, line 213: > To avoid conflicts or issues with other extensions, please put your name in the extension ID.

JeremyGamer13 (changes requested):

PR #597, opened 2026-06-21
breaking change: update GPU.sb3 to latest version

derpygamer2142:
see https://github.com/TurboWarp/extensions/pull/1762#issuecomment-4715729994
in summary: this is the best it's going to be.
penguinmod's version of this extension is severely outdated and every once in a while i still get a message about it. this update will probably break projects, i don't really remember. if you want i can make this v2 and unlist the old one, similar to how pen+ did it.
also updates the thumbnail because i was supposed to do that ages ago.
suggestion to anyone looking at this in the future: if you need compute shaders or pen+, maybe just use godot(or unity because i've heard their shader system is better right now)


jwklong:
looks like it wont break anything

PR #598, opened 2026-06-22
MMO Extension

Lawrence56354:


LordCat0:
1. An extension already exists with the same purpose of this
2. AI generated extensions aren't allowed

LordCat0:
also all this does is send a "join" event to the server, nothing else?

PR #604, opened 2026-07-06
black mold (slowed+reverb)

FloppyDisk-OSC:
added a natural decay
it can cover the screen
some optimization changes

it looks nicer imo


Ianyourgod:
leaving the extension running causes extreme lag, even after clearing then restarting the mold.

FloppyDisk-OSC:
aww i thought i fixed that

PR #605, opened 2026-07-10
re-organize extensions on the gallery

cicerorph:


src/lib/extensions.js
```diff
@@ -4,6 +4,7 @@
         to help copy & pasting work better
         pleas dont remove them :)
 */
+// a sakura
```
yuri-kiss, line 7: whar was this for

cicerorph: It was to check if Sakura Builder was working properly forgot to remove it

PR #606, opened 2026-07-10
My Extension

That-One-Developer:
This is penguin Windows! it is an extension I made that allows for opening windows and opening windows with HTML and closing windows as well. It can also track all windows available. It is a small extension, in both size and amount of blocks, but I think it will be useful


cicerorph:
You forgot to put the extension on lib/extensions.js

Also, read the Uploading Guidelines, creating a extension using AI is prohibited.

cicerorph:
Also a thing to notice, what if a project just decides to open 1000 windows at once? From what it seems there isn't a ratelimit or block for that.

That-One-Developer:
> Also a thing to notice, what if a project just decides to open 1000 windows at once? From what it seems there isn't a ratelimit or block for that.

alright going to fix my program for that. Thank you for informing me

PR #610, opened 2026-07-21
Fix bug in random utils

Dicuo:
changed the way the modulo stuff works so that negative seeds don't cause it to break

ignore the extra commits i'm gonna try to remove them somehow


Dicuo:
I can't remove the extra commits 💔
I remember last time this happened someone was able to remove it so uh do that please :))

RedMan13:
> I can't remove the extra commits 💔 I remember last time this happened someone was able to remove it so uh do that please :))

we cant do it, you need to revert the commits on your end

PR #612, opened 2026-07-28
Added Complex Numbers extension

salagata:
I added my first new extension Complex Numbers
Complex Number Type for do complex analysis functions.
Ideal for things were rotation is involved like bullet hells, or simulating things like fluids, illumination or camps


static/extensions/salagata/reisenComplex.js
```diff
@@ -0,0 +1,1387 @@
...
+         * Creates a complex number given it's polar form
+         *
+         * @static
+         * @param {number} absolute The absolute value, or modulus of the original number
+         * @param {number} argument The angle, in degrees
+         * @returns {ComplexNumberType} 
+         */
+        static fromPolar(absolute, argument) {
+            const real = absolute * Degrees.cos(argument)
+            const imaginary = absolute * Degrees.sin(argument)
```
Steve0Greatness, line 340: There work is pointless, as it gets discarded immediately when put into the constructor, which does this conversion on its own.

static/extensions/salagata/reisenComplex.js
```diff
@@ -0,0 +1,1387 @@
...
+         *
+         * @param {[number,number]|[number,number,number,number]} z Serialized 
+         * @returns {ComplexNumberType} 
+         */
+        Deserializer(z) {
+            if(z.length == 4) {
+                return new ComplexNumber.Type(z[0],z[1],z[2],z[3])
+            } else {
+                return new ComplexNumber.Type(z[0],z[1])
+            }
```
Steve0Greatness, line 385: This looks like a case where [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) would be more readable.

static/extensions/salagata/reisenComplex.js
```diff
@@ -0,0 +1,1387 @@
...
+                        blockType: Scratch.BlockType.REPORTER
+                    },
+                    {
+                        opcode: "complexToPolar",
+                        text: this.formatMessage("[COMPLEX] to polar form"),
+                        arguments: {
+                            COMPLEX: ComplexNumber.Argument
+                        },
+                        blockType: Scratch.BlockType.REPORTER
+                    },
```
Steve0Greatness, line 599: The should probably specify in the block name that they get stringified. Could also be combined into one block with a menu.

static/extensions/salagata/reisenComplex.js
Steve0Greatness: This seemingly happens quite a lot in this extension when working with polar form.

static/extensions/salagata/reisenComplex.js
```diff
@@ -0,0 +1,1387 @@
...
+         *
+         * @static
+         * @param {number} absolute The absolute value, or modulus of the original number
+         * @param {number} argument The angle, in degrees
+         * @returns {ComplexNumberType} 
+         */
+        static fromPolar(absolute, argument) {
+            const real = absolute * Degrees.cos(argument)
+            const imaginary = absolute * Degrees.sin(argument)
+            return new ComplexNumberType(real, imaginary, absolute, clampAngleDegrees(argument))
```
Steve0Greatness, line 341: Not particularly important, but could you be a bit more consistent with your choice of terminology: argument vs. phase.

static/extensions/salagata/reisenComplex.js
```diff
@@ -0,0 +1,1387 @@
...
+                        acceptReporters: false,
+                        items: [
+                            {
+                                text: this.formatMessage("positive"),
+                                value: "positive"
+                            },
+                            {
+                                text: this.formatMessage("negative"),
+                                value: "negative"
+                            },
```
Steve0Greatness, line 843: $\sqrt{b^2-4ac} > -b$ (IE sometimes both roots can be negative or positive) isn't always necessarily true, so labeling these as being "negative" or "positive" doesn't make much sense. It makes even less sense when applied to the complex plane.

There really isn't a satisfactory way to label roots (because they're unordered), so I guess it makes sense you'd do this.

static/extensions/salagata/reisenComplex.js
```diff
@@ -0,0 +1,1387 @@
...
+                
+                const phi = (0 + d * 360) / a;
+
+            return new ComplexNumberType(
+                    r * Degrees.cos(phi),
+                    r * Degrees.sin(phi),
+                    r, phi
+                );
+        }
+    }
```
Steve0Greatness, line 1385: Neither of these define `d` before using it.

static/extensions/salagata/reisenComplex.js
```diff
@@ -0,0 +1,1387 @@
...
+                            B: {
+                                type: Scratch.ArgumentType.NUMBER,
+                                defaultValue: 2
+                            },
+                        },
+                        ...ComplexNumber.Block
+                    },
+                    {
+                        opcode: "squareRoot",
+                        text: this.formatMessage("squareroot of [A]"),
```
Steve0Greatness, line 634: either `square root` to `sqrt` but not `squareroot`.

static/images/salagata/reisenComplex_placeholder.svg
Steve0Greatness, line 1: The contrast on this is *far* too low: <https://www.audioeye.com/color-contrast-checker/?foreground_color=D0D025&background_color=FFFE2E>. Use a darker color, for instance, black: <https://www.audioeye.com/color-contrast-checker/?foreground_color=000000&background_color=FFFE2E>.

static/extensions/salagata/reisenComplex.js
```diff
@@ -0,0 +1,1387 @@
...
+                //     return block.text
+                // })
+            }
+        }
+
+        getInfo() {
+            return {
+                id: "reisenComplexNumber",
+                name: this.formatMessage("Complex Numbers"),
+                description: this.formatMessage("Complex Number Type for do complex analysis functions, better implementation than the one made by jwklong in Mathemathics extension lmao"),
```
Steve0Greatness, line 421: I don't think a "description" key exists.

static/extensions/salagata/reisenComplex.js
```diff
@@ -0,0 +1,1387 @@
...
+
+            const r = subRadical == 2 ? Math.sqrt(A.absolute) : (A.absolute ** (1/subRadical));
+            const phi = A.argument / subRadical;
+
+            return new ComplexNumberType(
+                r * Degrees.cos(phi),
+                r * Degrees.sin(phi),
+                r, phi
+            );
+        }
```
Steve0Greatness, line 1125: You very much *can* do complex powers:

* assume you have complex numbers $a+bi$ and $c+di$
* $e^{\ln(a+bi)} = a+bi$
* $\therefore (a+bi)^{c+di} = e^{\ln(a+bi)^{c+di}} = e^{(c+di)\ln(a+bi)}$
* since every complex number as a polar form, and the polar form can be used to find the natural logarithm: $a+bi = r \angle \phi \implies \ln(a+bi) = r+\phi i$ (where $\phi$ is in radians)
* $e^{(c+di)\ln(a+bi)} = e^{(r+\phi i)(c+di)} = e^{rc-d\phi+(c\phi+rd)i} = e^{rc-d\phi}e^{(c\phi+rd)i}$
* $\therefore (a+bi)^{c+di} = e^{rc-d\phi}\angle(c\phi+rd)$ (in radians)

static/extensions/salagata/reisenComplex.js
```diff
@@ -0,0 +1,1387 @@
...
+
+            if(det ==  0) {
+                const unique = ( -b ) / (2 * a);
+                
+                solutions = [ new ComplexNumberType(unique,0), new ComplexNumberType(unique,0) ];
+            }
+            
+            if(det < 0) {
+                const positive = new ComplexNumberType(-b / (2 * a), Math.sqrt(Math.abs(det)) / (2 * a));
+                const negative = new ComplexNumberType(-b / (2 * a), Math.sqrt(Math.abs(det)) / (2 * a));
```
Steve0Greatness, line 1260: These are literally the same. I think you meant to add a minus sign infront of the complex part in the "negative" solution.

Steve0Greatness (changes requested):

static/extensions/salagata/reisenComplex.js
```diff
@@ -0,0 +1,1387 @@
...
+            const B = ComplexNumberType.toComplex(args.B);
+            const u = B.real ** 2 + B.imaginary ** 2;
+            
+            return new ComplexNumberType(
+                (A.real * B.real + A.imaginary * B.imaginary) / u, 
+                (A.imaginary * B.real - A.real * B.imaginary) / u
+                , A.absolute / B.absolute
+                , untransposeAngle(-(A.argument - B.argument))
+            );
+        }
```
Steve0Greatness, line 1034: A bit more of a general problem I have with this extension, but having 2 blocks that do, from an outside perspective, identical things is a really bad idea.

src/lib/extensions.js
```diff
@@ -321,6 +321,18 @@ export default [
         isGitHub: true,
         tags: ["new", "effects", "control", "data", "utility"]
     },
+    {
+        name: "Complex Numbers",
+        description: "Complex Number Type for do complex analysis functions, better implementation than the one made by jwklong in Mathemathics extension lmao.", 
+        code: "salagata/reisenComplex.js", 
+        banner: "salagata/reisenComplex_placeholder.svg",
+        creator: "salagata", 
+        tags: ["new","complex", "math", "graphics", "customtype", "utility"],
```
Steve0Greatness, line 330: Graphics? Please explain what this has to do with graphics.

src/lib/extensions.js
```diff
@@ -321,6 +321,18 @@ export default [
         isGitHub: true,
         tags: ["new", "effects", "control", "data", "utility"]
     },
+    {
+        name: "Complex Numbers",
+        description: "Complex Number Type for do complex analysis functions, better implementation than the one made by jwklong in Mathemathics extension lmao.",
```
Steve0Greatness, line 326: Descriptions are to describe your extension, for this, there's really no reason to mention another extension in your case.

src/lib/extensions.js
```diff
@@ -321,6 +321,18 @@ export default [
         isGitHub: true,
         tags: ["new", "effects", "control", "data", "utility"]
     },
+    {
+        name: "Complex Numbers",
+        description: "Complex Number Type for do complex analysis functions, better implementation than the one made by jwklong in Mathemathics extension lmao.", 
+        code: "salagata/reisenComplex.js", 
+        banner: "salagata/reisenComplex_placeholder.svg",
```
Steve0Greatness, line 328: Just about directory arrangement: since your extension is under your own directory, you don't need to add a prefix to the filename (note that the ID should still have a prefix, just that the filename doesn't need one, the directory structure already essentially does that for you).

src/lib/extensions.js
```diff
@@ -321,6 +321,18 @@ export default [
...
+        creator: "salagata", 
+        tags: ["new","complex", "math", "graphics", "customtype", "utility"], 
+        creatorAlias: "Reisen the Inaba", 
+        notes: "Additional help by jwklong extensions", 
+        unstable: false,
```
Steve0Greatness, line 333: It not being unstable is assumed by default, so you don't need to (and in fact probably shouldn't) specify it.

Steve0Greatness (changes requested):
I realize I haven't looked at the `extensions.js` changes yet, so here's a review there.

static/extensions/salagata/reisenComplex.js
salagata: One is supposed to operate with the real and imaginary part `divide()`, the other one is supposed to operate with the absolute value and phase without being subject to the floating-point error at the moment of calculating the real and imaginary part `divide2()`

Having two blocks is for operate separately what do you want to keep from the original complex number, or whether you're using the rectangular or polar form (indicated with "using the polar form").

Steve0Greatness: What I mean is that throughout this extension there's a division (mind the pun) between 2 entirely separate systems that don't have much impact on how people using the extension will use it. Having the extension have several different blocks that seem to perform the same operation with often minimal (if any) difference in how they appear can be very confusing.

I think it'd be better if you didn't make much of a distinction between complex numbers originally defined in polar form and numbers originally defined in the form $a+bi$. If you think the method using the absolute value and argument are more precise than doing just straight division, then maybe that should be the default way it gets done, rather than an alternate block that uses a secondary system.

static/extensions/salagata/reisenComplex.js
salagata: I think you're right in this one.
The constructor is supposed to receive 2 or 4 arguments, according the form of the number you're working with. 
If you pass the absolute value and phase, if you don't pass the real and imaginary part, it will calculate it, if not, then it will borrow the ones passed.
But It seems like I did the `fromPolar()` function without thinking on that, I will fix it when I can

static/extensions/salagata/reisenComplex.js
salagata: I thought here the code would be more explanatory but it seems like it might be a better idea to use the spread syntax.
Actually I was thinking to use spread syntax but...

static/extensions/salagata/reisenComplex.js
salagata: Great idea, I haven't thought on it. I will use a drop-down menu for this one.
Well, I thought to use terms like "stringified" would look rare for people who doesn't know what "to stringify" is. Maybe I can add "as text" at the end though.

static/extensions/salagata/reisenComplex.js
salagata: I will stay with phase, but my mind sometimes makes me change it with argument.
I'll try to be more consistent.

static/extensions/salagata/reisenComplex.js
salagata: I already know about the signs, but "Positive" and "Negative" are supposed to be the sign of the square root of the discriminant, the \sqrt{b^2-4ac} thing, the "Positive" and "Negative" are supposed to be the sign of \pm. Maybe i might change this for something... less confusing?

static/extensions/salagata/reisenComplex.js
salagata: Really sorry for this, this one was supposed to represent rational numbers more precisely but I was running out of time and I just did the things wrong here. When I tested this one, it had errors. I promise I will fix this one as quicker I can.

static/extensions/salagata/reisenComplex.js
salagata: OOPS, my mind sometimes crashes at trying to remember the way to spell... well, square root in English. First I think it is "square root", then I remember "sqrt" but I need the long one, then I incorrectly think it must be "squareroot".
I will fix this error.

static/images/salagata/reisenComplex_placeholder.svg
salagata: Ah. I was running out of time, so this banner is a placeholder I will change whether I'm free.
Thank you for the advice!

static/extensions/salagata/reisenComplex.js
salagata: I was also thinking on doing complex powers, but I was running out of time and I a haven't thought it might be useful, also I've thought people can implement their own using the blocks that we already have.

static/extensions/salagata/reisenComplex.js
salagata: Yes it is?. And, it is what is actually representing?
Why it looks like the same in your comment?

RedMan13: > Yes it is?. And, it is what is actually representing? Why it looks like the same in your comment?

what? no they are literally the same snippet copy-pasted, i know i dont know anything about complex numbers but i can certainly agree that `-1 != +1` and vice versa for any and all dimensions of numbers

src/lib/extensions.js
salagata: Things where rotations are involved. Since complex numbers are perfect for rotations, rotating a complex number is just multiplying a polar number with an absolute value of 1 (r=1) and with a defined angle (\theta), instead of rotating vectors, they use A Matrix for rotate. The complex numbers are better for this kind of things involving rotation in graphics.

src/lib/extensions.js
salagata: ... sorry I just ... N-NVM forget I did that-

salagata:
> I realize I haven't looked at the `extensions.js` changes yet, so here's a review there.

Thanks for the advice, I will consider all of them for the next time I'm free

static/extensions/salagata/reisenComplex.js
Steve0Greatness: That's true. There is an e^ function in the functions block (operators, not this extension), but I'd expect the functionality to appear in an extension dedicated to complex mathematics.

static/extensions/salagata/reisenComplex.js
salagata: AH SORRY, I was looking at the wrong line, I will fix it as soon as I can...

Steve0Greatness: > And, it is what is actually representing?

I think you already know that the determinate needs to either be added or subtracted, in this case both are adding the determinate, resulting in the same value being gotten for both roots, rather than different ones.

static/images/salagata/reisenComplex_placeholder.svg
Steve0Greatness, line 1: One last quick little note: extension thumbnails should be designed for a 2:1 aspect ratio (IE if you have 30 pixels in the width, your height would be 15), this current thumbnail isn't in that aspect ratio, so it doesn't quite appear correctly as of right now.

Screenshot of preview (see above comment from the sakurabuilder bot) showing how the thumbnail has a bit chopped off from the top and bottom

Notice how the top and bottom of the image are cut off. I'm aware this image is meant as a placeholder, but it's just a note for when you have a final thumbnail.

static/extensions/salagata/reisenComplex.js
salagata: SORRY I COMPLETELY KNOW, I just got a bit confused?, I will fix it as quicker I can

Steve0Greatness: Sorry, I hadn't seen your reply to that while I was writing my own reply (GitHub for some reason doesn't load new review comments until you reload the page). I didn't mean to rush you.

static/extensions/salagata/reisenComplex.js
salagata: I will see if i do an alternative system, i.e. , maybe choosing the mode when you want to do the, i.e. division.
As I said, sometimes it can't be uncertain if people will use the polar or rectangular form, i.e., in programs where the polar form is often better for calculate, i.e. rotations. Or in graphs, where the rectangular form is better.

src/lib/extensions.js
salagata: I will see if i do an alternative system, i.e. , maybe choosing the mode when you want to do the, i.e. division.
As I said, sometimes it can't be uncertain if people will use the polar or rectangular form, i.e., in programs where the polar form is often better for calculate, i.e. rotations. Or in graphs, where the rectangular form is better

static/extensions/salagata/reisenComplex.js
Steve0Greatness: Alright.

static/extensions/salagata/reisenComplex.js
yuri-kiss: > The should probably specify in the block name that they get stringified. Could also be combined into one block with a menu.

almost no extension that converts stuff to strings specifies this

Steve0Greatness: Actually, this is common on first-class type extensions like these.

yuri-kiss: i have literally never seen any extension do such a thing

yuri-kiss: not a single first class type extension that returns strings for values says the return type of string in its name besides the very obvious no-shit stringify block that obviously going to say that

Steve0Greatness: jwArray, dogeiscutObject, agBuffer, dogeiscutSet.

I'm saying this block should specify that it stringifies, like these extensions.

Steve0Greatness: > not a single first class type extension that returns strings for values says the return type of string in its name besides the very obvious no-shit stringify block that obviously going to say that

Not a single extension does that except the ones that do...?

yuri-kiss: image
the one in the typeof should arguably return object and not string but it still returns string without saying anything, and it can be considered "first class"

yuri-kiss: not a single one says it, the only blocks that say their name are for the purpose of stringifying, no other block that returns strings that isnt for the express purpose of stringifying says its type

Steve0Greatness: > not a single one says it, the only blocks that say their name are for the purpose of stringifying, no other block that returns strings that isnt for the express purpose of stringifying says its type

The issue here is that these blocks *are* for the express purpose of stringifing. They're the _only_ blocks in this extension that return strings, so they should specify.

> ~image removed~
> the one in the typeof should arguably return object and not string but it still returns string without saying anything, and it can be considered "first class"

Well, yes. Obviously the extension only returns stringifies JSON isn't going to specify that it only returns stringified JSON.

yuri-kiss: > > not a single one says it, the only blocks that say their name are for the purpose of stringifying, no other block that returns strings that isnt for the express purpose of stringifying says its type
> 
> The issue here is that these blocks _are_ for the express purpose of stringifing. They're the _only_ blocks in this extension that return strings, so they should specify.
> 

converting != express "stringifing"

> > ~image removed~
> > the one in the typeof should arguably return object and not string but it still returns string without saying anything, and it can be considered "first class"
> 
> Well, yes. Obviously the extension only returns stringifies JSON isn't going to specify that it only returns stringified JSON.

thats not an excuse, your point that its common is just false

Steve0Greatness: > thats not an excuse, your point that its common is just false

You yourself in the image above showed that it *is* common. There's quite a lot of precedent, the meer precense of which, in my opinion, means it's common, even if it not "the most common."

Steve0Greatness: > converting != express \"stringifing\"

It is stringifing it, though. This extension uses a class that isn't a string by default and these blocks stringify it in a specific format.

yuri-kiss: > > thats not an excuse, your point that its common is just false
> 
> You yourself in the image above showed that it _is_ common. There's quite a lot of precedent, the meer precense of which, in my opinion, means it's common, even if it not "the most common."

its not, thats very few blocks out of thousands

Steve0Greatness: > its not, thats very few blocks out of thousands

Thousands of blocks? Weren't we just talking about first-class extensions, not all blocks?

In any case, the fact that this extension does all of its computing in a non-text format makes, to me at least, it stringifying something important information. The extension you showed before (Mathematics) does its computations by passing around stringified JSON, it's already a string, so the information that it's being stringified isn't important, in this case, however, it being stringified is a) the entire point of the block, and b) something very different from the way the rest of the extension is handled, therefore it should say that it's stringifying the complex number.

static/images/salagata/complex_placeholder.svg
Steve0Greatness, line 1: For sure an improvement, but I'd personally switch the foreground and background colors; either that or darken the background a bit more (either-or, not both).

src/lib/Documentation/SalagataComplex.md
```diff
@@ -0,0 +1,77 @@
...
+---
+
+## Angles
+
+While developing this extension, I found a problem in the way that Scratch represents the Angles (Internally, and in ANGLE inputs).
+![Representation of angles in Scratch](https://files.catbox.moe/7uz5t7.png)
+The angles basically start in 0° but 0° is the top (when it should be in the right), then the front is 90°(when it should be in the top), which it's confusing.
+The trigonometrical functions (even the ones defined in ([sin v] of ()) block), recieve as argument a different kind of angle.
+Internally, Scratch has to do `90 - SCRATCH_ANGLE` in order to perform any trigonometrical operation. It converts the Scratch angle recieved commonly as input into a Plane angle(The one used in Mathematics).
+![Complex Plane angles](https://files.catbox.moe/umf5i8.png)
```
Steve0Greatness: You uploaded the images here, so they should be used in place of these clearly temporary uploads.

static/images/salagata/complexDocs/complexPlaneAngles.png
Steve0Greatness, line 1: Be *very* careful with your usage of files you find anywhere online. This file seems to come from Wikipedia under the CC BY-SA 3.0 license: [File:Complex number illustration modarg.svg](https://commons.wikimedia.org/wiki/File:Complex_number_illustration_modarg.svg). Make sure to attribute this source in the documentation, not doing so is a violation of the image's license.

src/lib/Documentation/SalagataComplex.md
```diff
@@ -0,0 +1,77 @@
...
+
+These Complex numbers have the property to choose the representation you want to use. Either the polar form or rectangular form
+
+---
+
+## Angles
+
+While developing this extension, I found a problem in the way that Scratch represents the Angles (Internally, and in ANGLE inputs).
+![Representation of angles in Scratch](https://files.catbox.moe/7uz5t7.png)
+The angles basically start in 0° but 0° is the top (when it should be in the right), then the front is 90°(when it should be in the top), which it's confusing.
```
Steve0Greatness: Rephrase suggestion:

In Scratch, 90&deg; is pointing left, with 0&deg; pointing up, which is different from the common representation which is the opposite.

src/lib/Documentation/SalagataComplex.md
```diff
@@ -0,0 +1,77 @@
...
+```scratch
+set direction to (90) ::motion  // Arguments are in Scratch angles
+direction ::motion reporter // Returns in Scratch angles
+[sin v] of (60) ::operators reporter // Arguments are in Complex Plane angles
+```
+
+In general.
+```
+COMPLEX_PLANE_ANGLE = 90 - SCRATCH_ANGLE
+SCRATCH_ANGLE = -COMPLEX_PLANE_ANGLE + 90
```
Steve0Greatness: Generally, when negative numbers and addition are involved the same expression, it's often cleaner and more understandable to represent that relationship with subtraction.

SCRATCH_ANGLE = 90 - COMPLEX_PLANE_ANGLE

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,1943 @@
...
+                docsURI
```
Steve0Greatness: You seem to have forgotten the documentation URL

Steve0Greatness (changes requested):
Couple small things.

salagata:
please?

static/extensions/salagata/complex.js
salagata: it's penguinmod docs URI, how do i include it?

salagata:
sorry for not answering in a while, i was busy and github didn't show me the comments

static/images/salagata/complexDocs/complexPlaneAngles.png
salagata: waaa how do i do it

Steve0Greatness: How you attribute? You just say where you got it and who made it.

static/extensions/salagata/complex.js
Steve0Greatness: Take the name of the file that you put it as, remove the `.md`, then append it to `https://extensions.penguinmod.com/docs/`. That's how you know the URL.

static/images/salagata/complexDocs/complexPlaneAngles.png
Steve0Greatness: This should be done on the docs page, by the way.

Steve0Greatness: Wikimedia Commons has its own preferred attribution, though:

image
image

I'd personally recommend copying it as HTML so it renders the way seen above the box in MD.

Then just paste that under the image in your docs.

src/lib/Documentation/SalagataComplex.md
Steve0Greatness: Not really what I meant. I meant keeping the uploads here (on the gallery) and replacing the Catbox.moe files with them. But alright.

Steve0Greatness:
> sorry for not answering in a while, i was busy and github didn't show me the comments

It's alright. I don't expect people to be able to reply instantly.

src/lib/Documentation/SalagataComplex.md
yuri-kiss: 90 degrees points right tho, -90 is left

yuri-kiss: image
did u typo?

Steve0Greatness: True, my bad

salagata: ok

src/lib/Documentation/SalagataComplex.md
salagata: you should have said that in a first place, waa

salagata: > Not really what I meant. I meant keeping the uploads here (on the gallery) and replacing the Catbox.moe files with them. But alright.

but in what file should i put them
because catbox.moe embeds are not loading for me

salagata: should i make a new folder called complexDocs/ and inside put the files?. Because i tried to put them in the images/ folder but it gave an svelte scope error, something

Steve0Greatness: In the same location you put them in this PR before.

static/extensions/salagata/complex.js
salagata: llright

src/lib/Documentation/SalagataComplex.md
salagata: > In the same location you put them in this PR before.

I put them in static/images/salagata/complexDocs , but svelte didn't load them

salagata: when i did that, static/images/salagata/complexDocs, it returned an svelte scope error, something like that. i'm not sure if put them in the lib/Documentation for avoid this error.

salagata: i think i'm putting them in src/lib/Documentation

Steve0Greatness: Sorry for the late response. They do go in the directly you said before (/static/images/salagata/complexDocs), but on the server they're accessed through /images/salagata/complexDocs (without the /static/).

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,2169 @@
...
+
+    
+    /**
+     * A complex number
+     *
+     * @class ComplexNumberType
+     * @typedef {ComplexNumberType}
+     */
+    class ComplexNumberType {
+        customId = "reisenComplexNumber";
```
Steve0Greatness, line 277: This needs to match the value you set below for the serializer registration: `salagataComplexNumber`.

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,2169 @@
...
+         * Support for the Jwklong Array Handler!
+         *
+         * @returns {string} 
+         */
+        jwArrayHandler() {
+            if(this._fromPolar) {
+                return `Complex<${this.real},${this.imaginary},${this.modulus},${this.phase}>`;
+            } else {
+                return `Complex<${this.real},${this.imaginary}>`;
+            }
```
Steve0Greatness, line 397: Both of these should really only show 2 values, you could probably delineate them in the same way you do below (in toString()).

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,2169 @@
...
+     * Transform an Complex Plane Angle into Scratch Angle
+     *
+     * @param {number} angle Complex Plane Angle
+     * @returns {number} 
+     */
+    function untransformAngle(angle) {
+        return -(angle - 90)
+    }
+
+    function standarizeJSONObject(json) {
```
Steve0Greatness, line 193: This function goes unused, as of currently.

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,2169 @@
...
+
+    /**
+     * Transform an Complex Plane Angle into Scratch Angle
+     *
+     * @param {number} angle Complex Plane Angle
+     * @returns {number} 
+     */
+    function untransformAngle(angle) {
+        return -(angle - 90)
+    }
```
Steve0Greatness, line 191: These functions are identical when simplified:

$-(\phi - 90) = 90 - \phi$

$-\phi + 90 = 90 - \phi$

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,2169 @@
...
+            "transform [ANGLE] into Complex Plane Angle": "transformar [ANGLE] en ángulo del plano complejo",
+            "transform [ANGLE] into Scratch Angle": "transformar [ANGLE] en ángulo de Scratch",
+        },
+    });
+        
+    const integrationsEnabled = {
+        jwVector: false,
+        jwArray: false,
+        dogeiscutSet: false,
+    }
```
Steve0Greatness, line 79: Unused.

Steve0Greatness (changes requested):

static/extensions/salagata/complex.js
salagata: ok

salagata: well i didn't wanted to mess up something, but okay

salagata:
done, hope it's not too late

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,2175 @@
...
+        return a
+    }
+
+    function castAngle(angle) {
+        return String((Math.sign(angle) === -1) ? angle + 360 : angle);
+    }
+
+    function isInteger(n) {
+        return Math.round(n) === n;
+    }
```
Steve0Greatness, line 171: 3 more unused chunks of code

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,2175 @@
...
+         * @param {number} [phase] Angle, in case you're working with polars, in degrees, transposed to the complex plane
+         */
+        constructor(real = 0,imaginary = 0, modulus, phase) {
+            // console.log(modulus, angle)
+            if(Boolean(modulus) || Boolean(phase)) {
+                this._fromPolar = true;
+                this._modulus = modulus;
+                this._phase = clampAngleDegrees(phase);
+                
+                switch (constrainAngleClampedPositive(this._phase)) {
```
Steve0Greatness, line 296: At this point, constrainAngleClampedPositive() and constrainAnglePositive() are probably unnecessary. This is the only use and it's entirely redundant due to constrainAnglePositive() doing the same exact thing as constrainAnglePositive(), and also covering the entire use-case of constrainAngleClampedPositive(), which wouldn't even work in a real world scenario due to it not modulating the angle, meaning its operation doesn't even work since any negative number which follows x < -360 would still be negative, and it allows positive values to grow infinitely.

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,2175 @@
...
+     * @returns {string}
+     */
+    function formatNumber(x) {
+        if (x >= 1e6) {
+            return x.toExponential(4)
+        } else {
+            x = Math.floor(x * 1000) / 1000
+            return x.toFixed(Math.min(3, (String(x).split('.')[1] || '').length))
+        }
+    }
```
Steve0Greatness, line 128: Unused.

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,2175 @@
...
+            Scratch.vm.runtime.registerSerializer(
+                "salagataComplexNumber",
+                ComplexNumber.Serializer, ComplexNumber.Deserializer
+            )
+
+            this.formatMessage = function (id) {
+                return Scratch.translate({ id: id, default: id });
+            };
+            
+            this.formatEveryBlock = function (blocks) {
```
Steve0Greatness, line 570: Dead code.

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,2175 @@
...
+                    case 270:
+                        this.real = 0;
+                        this.imaginary = -modulus;
+                        break;
+
+                    default:
+                        
+                        this.real = (isNaN(real) || real == 0) ? modulus * Degrees.cos(phase) : real;
+                        this.imaginary = (isNaN(imaginary) || imaginary == 0) ? modulus * Degrees.sin(phase) : imaginary;
+                        break;
```
Steve0Greatness, line 322: This switch statement might be better off if implemented in the `Degrees.cos` and `Degrees.sin` functions directly, as it means that other functions throughout the extension would be able to take advantage of this rather than just this constructor. Plus it'd clean it up a bit.

static/extensions/salagata/complex.js
```diff
@@ -0,0 +1,2175 @@
...
+        /**
+         * Deserializer for this type
+         *
+         * @param {[number,number]|[number,number,number,number]} z Serialized 
+         * @returns {ComplexNumberType} 
+         */
+        Deserializer(z) {
+            if(s[3]) {
+                s[3] = transformAngle(s[3]);
+            }
```
Steve0Greatness, line 535: Wrong input name. You're using z not s.

Steve0Greatness (changes requested):

static/extensions/salagata/complex.js
Steve0Greatness: To specify, I'm saying .formatEveryBlock() is dead code. It's unused.

PR #621, opened 2026-08-11
hotfix: fix mem leak + mix blend mode bug | lighting extension

Gen1xLol:
yeah


RedMan13:
and you tested this and it works

Gen1xLol:
> and you tested this and it works

yes, as far as i tested, it should workj

PR #622, opened 2026-08-15
Delist More Types Extension

lego7set:
I think the more types extension has been largely superseded by the core object, array, and lambda extensions, as well as the set extension. The more types extension is hard to use, and a lot of blocks are repurposed for multiple types rather than being specialized to do one thing while not using the new type API, which makes blocks look really long and anti-visual-coding. I think the extension is quite chaotic and not very useful with all of the better, cleaner alternatives.

For these reasons, I would like to delist the more types extension.

### DIFF
--- a/src/lib/extensions.js
+++ b/src/lib/extensions.js
@@ -548,14 +548,14 @@ export default [
         creator: "WAYLIVES",
         isGitHub: false,
     },
-    {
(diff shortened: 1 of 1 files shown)


Steve0Greatness (approved):
Fair reasons.

PR #626, opened 2026-08-20
add animated loaders extension

Ikelene:
its just animated loaders idk what else to say


jwklong:
wheeere is the css from

Ikelene:
pma-ring is mine, others are from a mix of claude (the progress bars were modified by me a little) and css-loaders.com
users can also import any custom loader they want from sites like https://css-loaders.com/

jwklong:
wouldnt custom loaders mean a xss vuln

Ikelene:
idk, should i remove it then? it kinda does let you run custom CSS so i would guess yes

Ikelene:
there fixed should be good for merge now if theres nothing else

Steve0Greatness:
I'm not actually opposed to allowing custom CSS, as long as there's some sort of sandboxing for custom loaders (like maybe having it in an iframe with some styling to hide that fact).

Ikelene:
> I'm not actually opposed to allowing custom CSS, as long as there's some sort of sandboxing for custom loaders (like maybe having it in an iframe with some styling to hide that fact).

I didn't actually think of that, that's a good idea. I think it's definitely something I'll toy with later, jwklong just didn't answer me for a bit so I just thought I'd remove it so I don't have to wait another 5 business days for another response.

static/extensions/Ikelene/loaders.js
```diff
@@ -0,0 +1,530 @@
...
+        }
+        this.container = container;
+      }
+
+      if (container.parentElement !== parent) {
+        parent.appendChild(container);
+      }
+
+      return container;
+    }
```
RedMan13, line 351: you should be using the built in stage overlays instead of attempting to find and overlay the stage, see https://github.com/PenguinMod/PenguinMod-Render/blob/89a587aff84e0a62daf907fba464d7e02f995027/src/RenderWebGL.js#L557 (addOverlay, removeOverlay)

static/extensions/Ikelene/loaders.js
```diff
@@ -0,0 +1,530 @@
...
+    _getOrNull(id) {
+      return this.loaders.get(String(id)) || null;
+    }
+
+    createLoader(args) {
+      const id = String(args.ID);
+      const container = this._ensureContainer();
+      if (!container) return;
+
+      let loader = this.loaders.get(id);
```
RedMan13, line 405: why do only these not use getOrNull?

static/extensions/Ikelene/loaders.js
```diff
@@ -0,0 +1,530 @@
...
+      loader.speed = speed;
+      loader.wrapper.style.setProperty("--pma-speed", String(speed));
+    }
+
+    setLoaderColor(args) {
+      const loader = this._getOrNull(args.ID);
+      if (!loader) return;
+      const hex = String(args.COLOR);
+      loader.color = hex;
+      loader.wrapper.style.setProperty("--pma-color", hex);
```
RedMan13, line 498: this isnt ever checked if its a color, any text can be put here making it xss

static/extensions/Ikelene/loaders.js
```diff
@@ -0,0 +1,530 @@
...
+      loader.style = String(args.STYLE);
+      loader.wrapper.style.setProperty("--pma-color", loader.color);
+      loader.wrapper.style.setProperty("--pma-speed", String(loader.speed));
+      loader.wrapper.style.display = loader.visible ? "" : "none";
+      this._renderPreset(loader);
+      this._applyTransform(id);
+    }
+
+    removeLoader(args) {
+      const id = String(args.ID);
```
RedMan13, line 433: i wont do this five trillion times cause im lazy
but this is simply wrong, you should be using Scratch.Cast.toString and Scratch.Cast.toNumber
especially critical for numbers because scratch explicitly will not ever cast to nan, and will even remove nan if given as a real number

static/extensions/Ikelene/loaders.js
RedMan13: you dont need to do this entirely yourself, `Scratch.Color.rgbToHex(Scratch.Cast.toRgbColorObject())` will plenty suffice

static/extensions/Ikelene/loaders.js
```diff
@@ -0,0 +1,530 @@
...
+      this._applyTransform(String(args.ID));
+    }
+
+    loaderExists(args) {
+      return this.loaders.has(String(args.ID));
+    }
+
+    loaderVisible(args) {
+      const loader = this._getOrNull(args.ID);
+      return !!loader && loader.visible;
```
RedMan13, line 515: why not just check if display is none

RedMan13 (changes requested):
i cant think of anything more so lets goo

PR #628, opened 2026-08-22
HTML Documents Extension

ScrTwPm:
Create HTML webpages and display them on the stage
htmldocs
I'll make a documentation when I know what you think of this


ScrTwPm:
But you were right about the Python extension it didnt really add anything new. I think this does add more features that that does. I'm checking right now to see if something like this already exists but i'm pretty sure it isn't

Steve0Greatness:
Doesn't appear to scale appropriately for full screened usage

image

Additionally, although the extension still believes the currently displayed page is still the page when changing view modes (ie. into/out of full screen), it does actually get removed.

image

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1083 @@
...
+            if(/[a-z0-9]/i.test(args.PAGE)){
+                this.pages[args.PAGE] = {
+                    "data":{
+                        "x":5,
+                        "y":5,
+                        height: "350",
+                        width: "470",
+                    },
+                    "ids": [],
+                    "code":""
```
Steve0Greatness: This would probably require a bit of a restructure, but there are far safer and more efficient ways to handle HTML code. Storing it as a DOM would actually be a good idea, here.

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1083 @@
...
+            "c 0 2 0 8 4 8 c 6 0 10 -7 14 -5 c 4 -2 8 5 14 5 c 4 0 4 -6 4 -8"
+         )
+        });
+}
+
+
+  class HTMLtoCanvas {
+    constructor(runtime) {
+      // Initialize an array holding your default dropdown menu options
+      this.pages = {}
```
Steve0Greatness: You'll probably want to use a `Map` instead of just an object, since maps are more optimized to change frequently.

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1083 @@
...
+  if (Scratch.gui) {
+    Scratch.gui.getBlockly().then(ScratchBlocks => {
+         ScratchBlocks.BlockSvg.registerCustomNotch("htmldocuments-coolshape", 
+            "c 2 0 3 1 4 2 l 4 4 c 1 1 4 -7 4 2 h 3 c 1 -1 2 -2 3 -3 c 1 1 2 2 3 3 l 3 0 c 0 -9 3 -1 4 -2 l 4 -4 c 1 -1 2 -2 4 -2"
+         )
+         ScratchBlocks.BlockSvg.registerCustomNotch("htmldocuments-coolshape2", 
+            "c 0 2 0 8 4 8 c 6 0 10 -7 14 -5 c 4 -2 8 5 14 5 c 4 0 4 -6 4 -8"
+         )
+        });
+}
```
Steve0Greatness: These currently go unused.

Steve0Greatness:
> Doesn't appear to scale appropriately for full screened usage
> ~image removed~

I see you had added a block that helps to scale appropriately (id `spw`). Second one still applies, though.

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1083 @@
...
+        }
+    }
+
+
+    displayPage(args, util){
+        try{
+            const elements = document.querySelectorAll(`#htmlpage`);
+            elements.forEach(el => el.remove());
+            const el = document.createElement("div");
+            el.innerHTML = this.pages[args.PAGE].code;
```
Steve0Greatness: For the purposes of sandboxing, please use [`iframe`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe) instead of a `div`. You can include HTML inside an `iframe` using the [`srcdoc`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#srcdoc) attribute.

Steve0Greatness:
Could you make it so that multiple pages can be displayed at once?

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1083 @@
...
+                        
+                        // Drag & Drop
+                        "dragstart", "drag", "dragenter", "dragover", "dragleave", "drop", "dragend",
+                        
+                        // Mobile & Media
+                        "touchstart", "touchmove", "touchend", "play"
+                    ]
+	        },
+            attr: {
+                acceptReporters: false,
```
Steve0Greatness: This should probably be set to true, instead of false. This would necessitate the use of a DOM, however, as the way you're currently setting attributes by their name is unsafe if you allow direct input.

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1083 @@
...
+            }
+        },
+
+		 { blockType: Scratch.BlockType.LABEL, text: "Event Listeners" },
+
+
+        		            {
+            opcode: 'eve',
+            blockType: Scratch.BlockType.HAT,
+            text: 'When listener for [ID] activated in [PAGE]',
```
Steve0Greatness: For consistency, this should start with a lowercase letter.

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1083 @@
...
+            } else {
+                value = args.VALUE
+            }
+               
+                // if (parentBlock.stylePage && parentBlock) {
+                    let page = args.PAGE//parentBlock.stylePage.page
+                    console.log("Before:", this.pages[page].code)
+                    this.pages[args.PAGE].code = `${this.pages[args.PAGE].code}<!--begin my style--><style>`
+                    this.pages[page].code = `${this.pages[page].code}.htmldocumentelement${args.TYPE}element${page}${args.NAME}{${args.PROPERTY}:${value}};`
+                    this.pages[args.PAGE].code = `${this.pages[args.PAGE].code}</style><!--end my style-->`
```
Steve0Greatness: In cases like these, it's easier to read if you use ``+= `...` `` instead of ``= `${...}...` ``

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1083 @@
...
+    }
+
+    resetDefault(args, util){
+        if(Object.keys(this.pages).includes(args.PAGE)){
+            this.pages[args.PAGE].data.width = Number(window.getComputedStyle(Scratch.renderer.canvas.parentElement).width.replace("px", "")) - 10
+            this.pages[args.PAGE].data.height = Number(window.getComputedStyle(Scratch.renderer.canvas.parentElement).height.replace("px", "")) - 10
+            this.pages[args.PAGE].data.x = 5
+            this.pages[args.PAGE].data.y = 5
+            console.log(this.viewing)
+            console.log(document.querySelector("#htmlpage"))
```
Steve0Greatness: There are a few spare console logs here and there in this extension, I'd recommend removing them before this gets merged.

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1083 @@
...
+                    }
+
+                    // Compare and trigger
+                    if (hatValue === triggerText && hatValuea === triggerTexta) {
+                        vm.runtime._pushThread(rootBlockId, target);
+                    }
+                }
+            });
+        });
+    });
```
Steve0Greatness: `runtime.startHats` has the functionality to start hat blocks dependent on their values, already.

Steve0Greatness (changes requested):
A bit more of a general concern, but you should put this through Prettier, because this is formatted, in my opinion, a bit badly.

ScrTwPm:
thanks for the feedback! i'll get to it.

Steve0Greatness:
> But you were right about the Python extension it didnt really add anything new. I think this does add more features that that does. I'm checking right now to see if something like this already exists but i'm pretty sure it isn't

Well, there's Project Interfaces, which is a similar idea, but it doesn't allow for the creation of documents made up of HTML.

ScrTwPm:
i was thinking about that exact thing while i was making this. I wanted to give users more control and the ability to display and build their own webpage documents.

ScrTwPm:
can i use 'https://unpkg.com/html-dom-parser@latest/dist/html-dom-parser.min.js' for prettier

ScrTwPm:
or i can use this function

```js
function prettierInText(html) {
  // Clean up existing whitespace and collapse it
  let clean = html.replace(/\s*([<>])\s*/g, '$1').replace(/\s+/g, ' ');
  let reg = /(<[^>]+>)/g;
  let matches = clean.split(reg).filter(Boolean);
  
  let formatted = '';
  let pad = 0;

  matches.forEach((token) => {
    // Check if the token is a closing tag
    if (token.match( /<\/\w+/ )) {
      pad--;
    }
    
    // Add current line indentation
    formatted += '  '.repeat(Math.max(0, pad)) + token + '\n';
    
    // Check if the token is an opening tag (and not self-closing)
    if (token.match( /<[^\/][^>]*[^>\/]>/ ) && !token.match(/<(input|img|br|hr|meta|link)/)) {
      pad++;
    }
  });

  return formatted.trim();
}

// Example Usage
const raw = '<div><h1>Title</h1><p>Text</p><img src="img.jpg"/></div>';
console.log(prettierInText(raw));
```

Gen1xLol:
> or i can use this function
> 
> ```js
> function prettierInText(html) {
>   // Clean up existing whitespace and collapse it
>   let clean = html.replace(/\s*([<>])\s*/g, '$1').replace(/\s+/g, ' ');
>   let reg = /(<[^>]+>)/g;
>   let matches = clean.split(reg).filter(Boolean);
>   
>   let formatted = '';
>   let pad = 0;
> 
>   matches.forEach((token) => {
>     // Check if the token is a closing tag
>     if (token.match( /<\/\w+/ )) {
>       pad--;
>     }
>     
>     // Add current line indentation
>     formatted += '  '.repeat(Math.max(0, pad)) + token + '\n';
>     
>     // Check if the token is an opening tag (and not self-closing)
>     if (token.match( /<[^\/][^>]*[^>\/]>/ ) && !token.match(/<(input|img|br|hr|meta|link)/)) {
>       pad++;
>     }
>   });
> 
>   return formatted.trim();
> }
> 
> // Example Usage
> const raw = '<div><h1>Title</h1><p>Text</p></div>';
> console.log(prettierInText(raw));
> ```

This seems to be very clearly AI generated. I don't know if you vibecoded this extension but if you did that's a biiiig no-no

ScrTwPm:
I typed this myself. But the function i just pasted was AI generated because I wanted to see if you would or would not want me to import the prettier.
I typed the full extension by hand. I've been working on it since May.

ScrTwPm:
I don't paln on using the script I sent you but I want to verify if i'm ok to import perttier from https://unpkg.com/html-dom-parser@latest/dist/html-dom-parser.min.js'

ScrTwPm:
> This seems to be very clearly AI generated. I don't know if you vibecoded this extension but if you did that's a biiiig no-no

I built this off of my previous html to canvas extension that I made last year.
It is very unsafe and uses cst1229's extendable blocks (i never made a pr for it)

[htmltocanvasold.js](https://github.com/user-attachments/files/31339949/htmltocanvasold.js)

Steve0Greatness:
> I don't paln on using the script I sent you but I want to verify if i'm ok to import perttier from https://unpkg.com/html-dom-parser@latest/dist/html-dom-parser.min.js'

Uh, I don't think that's Prettier, that's a DOM parser. [Prettier](https://prettier.io/) is a code formatter, I was requesting that you reformat your code.

ScrTwPm:
OH. ok. it does have a html parser. You wnated me to format my code

Steve0Greatness:
> OH. ok. it does have a html parser. You wnated me to format my code

Yes. The JavaScript of the extension, that's what I want you to format.

ScrTwPm:
I'll do that when i post my new revided version that replies to your suggestions

(I'm a good coder but a horrible formatter. I'm downloading prettier on vscode right now)

Steve0Greatness:
> I'll do that when i post my new revided version that replies to your suggestions
> 
> (I'm a good coder but a horrible formatter. I'm downloading prettier on vscode right now)

Alright, you actually don't need to open a new PR, just push to the same branch that this PR is pulling from (main).

Steve0Greatness:
Additionally, you can just use the builtin [`DOMParser`](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser) provided by the browser, rather than needing an external package..

ScrTwPm:
I got you.

ScrTwPm:
should i use that to insert elements into my iframe or change the entire 
```js
this.pages[page].code
```
to a HTML DOM

ScrTwPm:
New:
* got rid of unused custom notch shapes
* iframe instead of div
* lowercased when event listener...
* used += `...` instead of ... = `${...}...`
* removed spare console logging

Did not:
* use a DOM to store webpages (tell me if this is a need before it is merged)
* Change to a Map (please expound on what needs to be changed to Map)
* Change to runtime.startHats: (it didn't work for me when i tried to use it. Maybe I'm doing it wrong.)

ScrTwPm:
I'm in eastern time plz tell me what you think

ScrTwPm:
I'll be back on Monday evening

ScrTwPm:
going on a day trip

Steve0Greatness:
> should i use that to insert elements into my iframe or change the entire
> 
> ```js
> this.pages[page].code
> ```
> 
> to a HTML DOM

Ideally, the code property should be a DOM.

Steve0Greatness:
> * please expound on what needs to be changed to Map

I'm suggesting you use a map for `pages`. Maps are best at dynamic hashmaps that are anticipated to have new elements created and/or removed frequently.

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1116 @@
...
+                            "y": 5,
+                            height: "350",
+                            width: "470",
+                        },
+                        "ids": [],
+                        "code": ""
+                    }
+                } else {
+                    throw new Error("Name must contain at least one number or letter")
+                }
```
Steve0Greatness: This requirement disallows non-Latin alphabet languages from being used.

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1116 @@
...
+                        ]
+
+                    }
+
+                }
+            };
+        }
+
+        createPage(args, util) {
+            if (!Object.keys(this.pages).includes(args.PAGE)) {
```
Steve0Greatness: Create page should also reinitialize a page (IE, reset its x, y, width, and height). If it doesn't, it could result in issues when rerunning projects that don't create and immediately clear their newly created pages, since pages aren't cleared between runs of a project.

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,1116 @@
...
+
+        eve() {
+            return true;
+        }
+
+        current() {
+            if (document.querySelector(".htmlpage")) {
+                return (JSON.stringify(this.viewing))
+            } else {
+                return ('')
```
Steve0Greatness: should return `'[]'`, as this is expected to be an array.

Steve0Greatness (changes requested):

Steve0Greatness:
> going on a day trip

Alright. This'll just be waiting for when you get back.

ScrTwPm:
> Ideally, the code property should be a DOM.

what about nesting elements?

ScrTwPm:
Right now, non-self-closing elements open the element (in the string: i.e., ```<div>```) then close it (```</div>```)back once all the blocks inside of it have run (nesting elements). 

image

I'm not sure how I will change this just using a DOM. otherwise the idea is fine. Working on a solution right now

ScrTwPm:
WHYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYY a DOM????????????????????????????????????????????????????????????????????????/

ScrTwPm:
Used a map for ```pages``` and return [] on empty "current page displayed". Added support for non-latin languages by only rejecting empty names for pages.

RedMan13:
> WHYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYY a DOM????????????????????????????????????????????????????????????????????????/

The Document Object Model (DOM) is the system used to represent the contents of a Document as an Object. In the web, a document is normally some flavor of Standard Generalized Markup Language (SGML). The current standard for representing such documents as Objects, is to make each "Element" into one object, that object would then house each property of that Element, such as attributes and child Elements. Normally, this structure is installed in languages so that the Document can be controlled by, and better understood by, the developers of a program, and the program itself. There have been, in the past, propositions to standardize syntax for interacting with SGML inside of languages that often need to, such as JavaScript. Though JavaScript did not end up accepting this proposed standard.

Gen1xLol:
> WHYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYY a DOM????????????????????????????????????????????????????????????????????????/

I know this is unrelated to the issue and will most likely be marked as off topic but Why are you Like This

Steve0Greatness:
> WHYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYY a DOM????????????????????????????????????????????????????????????????????????/

Because a DOM is safer than your current method.

ScrTwPm:
Can i just change ```this.pages...code```` to a DOM after I perform the same string operations to it?
Or do I have to change my operations to edit the DOM rather than on the string

like using 
```js
// a function to add elements
docVar.createElement("div")
```
 or can I just use
```js
const toAString = new XMLSerializer()
const dom = new DOMParser()
// getinfo...
// a function to add elements
string = toAString.serializeToString(this.pages...code) //obviously this is a map now so different ops like get()
string += "<div>some stuff...</div>"
this.pages...code = dom.parseFromString(string, 'text/html') // set() now
```

Did you see my comment about how the nesting elements works. If this can be solved I can completely move to operations on the dom rather than on strings

Screenshot 2026-08-25 134023


thats why i asked 

> WHYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYY a DOM????????????????????????????????????????????????????????????????????????/

jwklong:
you can run unsandboxed javascript code with this extension
{0CFB0602-D871-4682-B118-2CA899034480}

ScrTwPm:
I'll disallow using <script> in the add text block

ScrTwPm:
@jwklong do you have a resolution for the comment I sent before your find?

jwklong:
you also cant make html code in parallel
{6F8E2FAC-ED08-400A-AB88-30BE0FD1E10E}

ScrTwPm:
uhh what should happen when that code is built? It does what I coded it to do. 
The loop at the bottom begins first and appends ```<div id="...new-el">```
Then before it is done the new thread begins and appends ```<div id ="...new-el2>``` and ```</div>```
After the 2nd second, ```</div>``` is added from the loop at the bottom. 
I don't know why anyone would need to build that script.
Screenshot 2026-08-25 134023

ScrTwPm:
I will make it do whatever you think it should do if someone does that. I dont know what a user would expect if they did that.

Steve0Greatness:
> I'm not sure how I will change this just using a DOM. otherwise the idea is fine. Working on a solution right now

Actually, I'd argue, in part related to Jwk's argument, that this is actually a place where a DOM is better. You can set values on the thread (`util.thread`) that will only be on the current thread, meaning multiple threads can create different elements at the same time. You could set a value that is an array of all ids in order of last created. New elements get the latest id, and when the element closes it `pop`s the current id.

Speaking of which, there should actually be a way to edit the child elements and text of an element after its creation.

ScrTwPm:
so the there is a thread var (an array) that contains elements to be created. i'm changing it right now to a dom

ScrTwPm:
changed to a DOM

ScrTwPm:
Added new blocks and got rid of unessecary tracking of ids

ScrTwPm:
Didn't mean to close :slappingmyheademoji:

static/extensions/ScrTwPm/htmldocuments.js
```diff
@@ -0,0 +1,2010 @@
...
...7FbM+XpTm8Z3TIwmZuRzwelT+2ZT9s411wDeKipxBccqXVZ517NLjo0i+rM3bchfIOAfcAnx+f6emOBXGeFBfszqzUgZ2kycXZscNnpgf23pDsvvv5/c99fEfypb9Tdnk/ItCbo4HZeFT+WCMtpVDnPZXaDd+MvvI665+385B3xpkfk1HjCyjv0XGY73Y15c6pcsHu5MEDtmLNWaTDs81a51uT3Xd/S/+mT21PvvhJkT18MyYZPH7wd9nn5idBbg02S8CkiK0X4l/6NGpXPYPohudZb8d5XZT/KYT4CHDr+HDmw25U7mwEF+xO3uS4dj4EtpImrzD7H35N99N/f8ng9i/56d7dmKP7sb2jlOpM2GzIZOFYhG1Y2IZsb0XvvILwG15AcMmVmWy270N5HwD+Gjgybm3bc10RnY3mgt0pigBrbjT7HnxL8uD9N3bvu6uR3n+Hb+65yc8evUMWboPU9PRK+wLrX/ZM9MVXGn3xFYl/0WWJd9Z5B2QYfRT4o/Go3FWvOLlwwe4UU5qcazqrrzLHDr4oPbB3R//WL25LvvjJxezhm33TOQS6PRoZ5zk3v76mXHmI7TuRl3+Dja5//qC2c9dBf/uOI7LRugvl/w+E+H+uI6IzLy7YnaITwA7S5EWms3rDcN8jZyb3335x/3MfP3/42b+rkxwdbZCazM9v1Gh+XU35ZNs+6QDvkm+g9qyX2/Ca67v6kivule2tu1H+Awjxd+PqFRfmzty5YHfKZgG4DrjGrC6fObjlC7sGX/7cdenD9203+x70Tqt2fn1NuReM+q9sOw91zkVWX3Z1HN3wvIe8sy/4MsrbA+wfB/lX3METTtG4YHfKTAPnApea1eXzho/sefbwofuem9x3x5nJHTd52UNfwg66gHfiRdjxoidmtH6pdlyGt/MauOhyG116RS/cuesrcnHbP6K8u8ZNtO4HDrit+06RuWB3qmJUUjnqUrnF7H/42uHhAy/v3nXbU4e3/UvT3Pslle2/G7waQtch62D7q4gzLsG74Bqrd15p9WVXJ965F+7xt+/4R9ls/xPKe2hcgnhgPCp37W2dUnDB7lSRGAd9gDV1s7pySba28q3p3gef3//0x3ckd98c2O3nEV3//DjcdfV+74wzPyfD6G9Q/i0IsToejSfAuCm945TL/w+xlwJr2wCYTgAAAABJRU5ErkJggg==',
```
Steve0Greatness: This is far to large for a menu image. Keep in mind the scale: they're usually displayed very small on the page. A more reasonable size would be around 32x32.

Steve0Greatness (changes requested):
I'll probably do a somewhat more in-depth review at some point, but I noticed you added a menu icon. It's currently 374x374px, which is far to large (note they're always displayed at very small resolutions). My personal recommendation is to scale it to 32x32.

ScrTwPm:
yeah i´ll do that.

ScrTwPm:
image

ScrTwPm:
```runtime.startHats``` finally worked!

**Comment by ScrTwPm** [comment]
I'll be back on Monday.

**Comment by ScrTwPm** [comment]
back

**Comment by Steve0Greatness** [comment]
There should probably be a way to change text content of an element dynamically.

**Comment by ScrTwPm** [comment]
This? 
image

**Comment by ScrTwPm** [comment]
or are you talking about changing the elements while the page is showing without having to display it again to reload the changes

**Comment by ScrTwPm** [comment]
image

**Comment by Steve0Greatness** [comment]
> or are you talking about changing the elements while the page is showing without having to display it again to reload the changes

Yes.

**Comment by ScrTwPm** [comment]
Done

**Comment by ScrTwPm** [comment]
all changes made

**Comment by RedMan13** [comment]
if you could please fix the merge conflicts github lists that would be perfect

**Comment by jwklong** [comment]
your custom types need serialization/deserialization support or projects may not be able to safe if for example a custom type is stored inside a variable

**Comment by ScrTwPm** [comment]
> if you could please fix the merge conflicts github lists that would be perfect

Can i make a new PR? this one has too many commits anyway.

**Comment by ScrTwPm** [comment]
> your custom types need serialization/deserialization support or projects may not be able to safe if for example a custom type is stored inside a variable

I'll get on it

**Comment by ScrTwPm** [comment]
ill finish the serilizzaton by the end of the week

**Comment by ScrTwPm** [comment]
added

**Comment by ScrTwPm** [comment]
hello?

File: static/extensions/ScrTwPm/htmldocuments.js [hunk 14]
```diff
@@ -43,9 +46,25 @@
         }
     }
 
-
+    const vms = Scratch.vm
+    console.log(vm)
```
RedMan13: whyr console dot logging, and a different var that's the same thing as global

RedMan13:
> hello?

sorry about that

ScrTwPm:
done

ScrTwPm:
...?

ScrTwPm:
are we good to go or not

Steve0Greatness:
Sorry that this kind of thing takes awhile. Jwk wants to review this before it gets merged but hasn't gotten around to it, yet.

jwklong:
xss vulnerability!!!! also its pretty concerning that i didnt have to display the page for the xss to function
{299AD03D-E954-4467-97CA-1D6344D75D86}

Steve0Greatness:
Looks like the add text block sets with document fragment rather than setting plain text. It should probably be setting plain text.

ScrTwPm:
done.

PR #631, opened 2026-08-30
iris text - yet another update

Gen1xLol:
removes apostrophes and quotation marks from the built-in punctuation list, adds gradients and updates the animate block to include "and wait/don't wait" (i tested it and projects that alr have it should work)


Gen1xLol:
imagen

yay

PR #640, opened 2026-09-05
midi update :glee:

FloppyDisk-OSC:
"is instrument drum?"
"when note released"
"current instrument number"

ok :3


FloppyDisk-OSC:
dont merge i fucked up the icon oh no

PR #643, opened 2026-09-06
Penguin AI

bta35628-cmd:


Dicuo:
> Your extension CANNOT rely on the use of any generative AI. This is mainly for extensions that respond to messages from AI like ChatGPT. Any extensions added before this rule are an exception.

